Issue #19, opened 2022-11-10
Sort out relative imports


matentzn:
In order to do testing, we need to be able to import `pheval.xyz` without having to resort to relative imports with a `.` etc.

souzadevinicius:
Fixed with some code adjustments in runner and we must run poetry shell before tox execution.

PR #21, opened 2022-11-11
fixing issue #19

souzadevinicius:


matentzn:
Can you link this PR to the ticket it solves, and explain why this suddenly works? Mysterious!

souzadevinicius:
The PR solves issue #19, and I have provided additional details about the solution in the issue close description.
Tox must be run inside the poetry shell because that's where the dependencies are installed.

PR #22, opened 2022-11-11
17 18 runner dependency injection

souzadevinicius:
This solution was implemented before we met with @pkalita-lbl. Dependency injection is performed using a small python library.
A parameter passed by the user in the PhEval CLI will automatically inject the runner dependency.
An example of the new help output can be found here

```
Usage: pheval [OPTIONS] COMMAND [ARGS]...

Options:
  -r, --runner RUNNER  Runner  [required]
  --help               Show this message and exit.

Commands:
  run
```
There are two valid runners implemented
- ExomiserRunner
- FooRunner

The "-r" flag describes the runner that will be used in the pipeline.
They have a concrete run implementation that only prints out the runner's name.

Valid run executions: 
- pheval -r exomiser run -i test
```
running pheval::run command
Running with exomiser runner
```
- pheval --runner foo -i test
```
running pheval::run command
Running with foo runner
```

Any other runner will throw an error from click

```
Try 'pheval --help' for help.

Error: Invalid value for '--runner' / '-r': 'bar' is not one of 'exomiser', 'foo'.
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -18,6 +18,23 @@ optional = false
 python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*,!=3.6.*,>=2.7"
 
 [[package]]
+name = "dependency-injector"
+version = "4.40.0"
+description = "Dependency injection framework for Python"
+category = "main"
+optional = false
+python-versions = "*"
+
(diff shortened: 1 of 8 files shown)


matentzn:
Can you provide an example of how this would work if you want to provide a runner that is not part of the package, i.e. defined externally?

PR #25, opened 2022-11-17
Runner setup

souzadevinicius:


src/pheval/cli_pheval.py
```diff
@@ -4,8 +4,40 @@
 
 import click
 
+from .pipeline import DefaultPhEvalRunner
+from .runners.exomiser_pheval_runner import ExomiserPhEvalRunner
```
matentzn, line 8: This is only temporary, until we figure out dependency injection with poetry

matentzn (approved):

PR #48, opened 2023-01-25
Visualising Semantic Similarity Profiles and their diffs

### LINKED ISSUE #43: Visualising Semantic Similarity Profiles and their diffs
### SEMSIM1:

| subject_id | object_id | jaccard |
| ------ | ----- | ------ |
| HP:001 | MP:009 | 0.5 |
| HP:002 | MP:008 | 0 |
| HP:003 | MP:007 | 0.8 |

### SEMSIM2:

| subject_id | object_id | jaccard |
| ------ | ----- | ------ |
| HP:001 | MP:009 | 0.5 |
| HP:002 | MP:008 | 0.3 |
| HP:003 | MP:007 | 0.1 |

### DIFF (SEMSIM1 - SEMSIM2)

| subject_id | object_id | jaccard |
| ------ | ----- | ------ |
| HP:001 | MP:009 | 0 |
| HP:002 | MP:008 | -0.3 |
| HP:003 | MP:007 | 0.7 |

### Visualisation ideas:

- [x] Heatmap for DIFF (ditching all 0s)
    - Group scores by top level ontology terms
- [ ] Aggregating averages up the ontology hierarchy, plotting a heatmap of the upper categories only

souzadevinicius:
Fixes #43


src/pheval/cli_pheval_utils.py
```diff
@@ -11,19 +14,77 @@
...
+    output_file_suffix: str,
+    output_dir: click.Path,
+):
+    """Generate noisy phenopackets from existing ones."""
+    create_scrambled_phenopackets(output_dir, output_file_suffix, phenopacket_dir, scramble_factor)
+
+
+@click.command("semsim-heatmap")
+@click.option(
+    "--semsim1",
```
matentzn, line 67: lets standardise diff parameters to `left`, `right`, so `-L/--semsim-left`, `-R/--semsim-right`, (make sure -R and -L are not taken and not used as standard parameters), see https://github.com/monarch-initiative/operations/issues/118

src/pheval/utils/file_utils.py
```diff
@@ -34,3 +38,41 @@ def obtain_closest_file_name(file_to_be_queried: Path, file_paths: list[Path]) -
...
+    """Ensures the existence of files passed as parameter
+    Raises:
+        FileNotFoundError: If any file passed as a parameter doesn't exist a FileNotFound Exception will be raised
+    """
+    for file in files:
+        if not path.isfile(file):
+            raise FileNotFoundError(f"File {file} not found")
+
+
+def ensure_columns_exists(**kwargs):
```
matentzn, line 53: this is not very readable - is kwargs really necessary? all arguments should have proper types.

src/pheval/utils/semsim_utils.py
```diff
@@ -0,0 +1,86 @@
...
+import click
+import numpy
+import pandas as pd
+import plotly.express as px
+
+import pheval.utils.file_utils as file_utils
+
+
+def filter_non_0_score(data: pd.DataFrame, col: str) -> pd.DataFrame:
+    """Gets rid of data with a score of 0 in the selected column
```
matentzn, line 15: Please document better - not sure what it does

matentzn (changes requested):
Will add more later, need to eat

src/pheval/cli_pheval_utils.py
```diff
@@ -11,19 +14,77 @@
...
+    required=True,
+    help="Path for creation of output directory",
+    default="noisy_phenopackets",
+)
+def scramble_phenopackets_command(
+    phenopacket_dir: click.Path,
+    scramble_factor: float,
+    output_file_suffix: str,
+    output_dir: click.Path,
+):
```
yaseminbridges, line 60: I think it would be better that if using the scramble-phenopackets command here for it to have all the click options defined as it is in the [create_noisy_phenopackets.py](https://github.com/monarch-initiative/pheval/blob/b05ae24572c39837b39d6237e559aa9df24e6722/src/pheval/prepare/create_noisy_phenopackets.py#L214). Specifically, where I have specified the option type as 'Path'.

src/pheval/cli_pheval_utils.py
```diff
@@ -11,19 +14,77 @@
...
+@click.option(
+    "--output-dir",
+    "-O",
+    metavar="PATH",
+    required=True,
+    help="Path for creation of output directory",
+    default="noisy_phenopackets",
+)
+def scramble_phenopackets_command(
+    phenopacket_dir: click.Path,
```
yaseminbridges, line 56: I am also not sure on the difference on the difference between click.Path and Path from pathlib - however I have defined all my paths from pathlib and not sure whether the functions will still perform the same when defining it as a click.Path. It would probably be best to stay consistent with defining Path types

src/pheval/utils/semsim_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Returns:
+        pd.Dataframe: parsed semantic similarity dataframe
+    """
+    df[cols[-1]] = pd.to_numeric(df[cols[-1]], errors="coerce")
+    df.replace("None", numpy.nan).dropna(subset=cols[-1], inplace=True)
+    return df
+
+
+def diff_semsim(data1: pd.DataFrame, data2: pd.DataFrame, score_column: str) -> pd.DataFrame:
```
yaseminbridges, line 42: I think it would be more useful to give variables more meaningful names, i.e. instead of data1 and data2 use semsim1 and semsim2

yaseminbridges (commented):
Looks good! I only had some minor comments

src/pheval/cli_pheval_utils.py
matentzn: We should always be using pathlib imo

src/pheval/cli_pheval_utils.py
matentzn: I would advocate for a clear separation of CLI methods and their implementation. I think cli_pheval_utils should contain all CLI mappings related to pheval utils, and the other modules (like [create_noisy_phenopackets.py](https://github.com/monarch-initiative/pheval/blob/b05ae24572c39837b39d6237e559aa9df24e6722/src/pheval/prepare/create_noisy_phenopackets.py#L214)) only their implementation. Are you advocating to keeping CLI methods in the same files as their implementation @yaseminbridges? 

We still need to work out the correct options - for example, we need to be able to pass in single phenopackets as well as entire corpora - work in progress.

yaseminbridges: No no, I agree that they should be clearly separated. My point was more that when I define the click command, for each option I am defining the type expected, for example [here](https://github.com/monarch-initiative/pheval/blob/b05ae24572c39837b39d6237e559aa9df24e6722/src/pheval/prepare/create_noisy_phenopackets.py#L221) where I've defined the type as the Path. 

There is already a separate command for creating a single noisy phenopacket [here](https://github.com/monarch-initiative/pheval/blob/b05ae24572c39837b39d6237e559aa9df24e6722/src/pheval/prepare/create_noisy_phenopackets.py#L204) which could also be put into this file

src/pheval/cli_pheval_utils.py
```diff
@@ -11,19 +16,120 @@
...
+    required=True,
+    metavar="FILE",
+    help="Path to the second semantic similarity profile.",
+)
+@click.option(
+    "--score-column",
+    "-s",
+    required=True,
+    type=str,
+    help="Score column that will be used in comparison (e.g jaccard_similarity)",
```
yaseminbridges, line 87: It may be useful to define this option type as a click.Choice - this way we can easily know which score columns are available as a comparison and can limit the chance of choosing an invalid choice

```python
@click.option('--score-column',
              type=click.Choice(['jaccard_similarity', 'other_option', 'other_option' ...]))

File: src/pheval/cli_pheval_utils.py [hunk 8]
```diff
@@ -11,19 +16,120 @@
...
+)
+@click.option(
+    "--analysis",
+    "-a",
+    default="heatmap",
+    type=str,
+    help="""There are two types of analysis:
```
yaseminbridges, line 94: You can use 
```python 
@click.option('--analysis',
              type=click.Choice(['heatmap', 'percentage_diff'], case_sensitive=False....))
```
here again, and this allows you to remove lines 125-130 as click will automatically throw an error if any of these choices are not selected

src/pheval/cli_pheval_utils.py
```diff
@@ -11,19 +16,120 @@
...
+        score column for this purpose. Defaults to "heatmap".
+        percentage_diff - Calculates the score column percentage difference between the semantic similarity profiles""",
+)
+@click.option(
+    "--output",
+    "-o",
     metavar="FILE",
-    help="Path to the phenopacket to be spiked.",
+    default="diff.semsim.tsv",
+    help="Output path for the difference tsv. Defaults to percentage_diff.semsim.tsv",
```
yaseminbridges, line 104: The default file name for this option is different to the one stated in the `help` and then later on in the docstring

yaseminbridges (commented):
Looks good! 😄 I just had a couple of minor comments

Makefile
```diff
@@ -0,0 +1,74 @@
...
+	wget $(URIBASE)/mp/mp-base.owl -O $@
+
+$(RAW_DATA_FOLDER)/hp-mp-merged.owl: $(RAW_DATA_FOLDER)/hp.owl $(RAW_DATA_FOLDER)/mp.owl
+	test -d $(RAW_DATA_FOLDER) || mkdir -p $(RAW_DATA_FOLDER)
+	robot merge --input $(RAW_DATA_FOLDER)/hp.owl --input $(RAW_DATA_FOLDER)/mp.owl reason reduce --output $@
+
+$(RAW_DATA_FOLDER)/hp-mp-merged2.owl: $(RAW_DATA_FOLDER)/hp-mp-merged.owl
+	test -d $(RAW_DATA_FOLDER) || mkdir -p $(RAW_DATA_FOLDER)
+	$(eval TERM1=$(shell bash -c "cat $(RAW_DATA_FOLDER)/random-hp-terms.txt | cut -d ' ' -f1 "))
+	$(eval TERM2=$(shell bash -c "cat $(RAW_DATA_FOLDER)/random-hp-terms.txt | cut -d ' ' -f2 "))
```
matentzn, line 39:
```suggestion
	$(eval TERM2=$(shell bash -c "cat $(RAW_DATA_FOLDER)/random-mp-terms.txt | cut -d ' ' -f2 "))
```

Makefile
```diff
@@ -0,0 +1,74 @@
...
+	test -d $(RAW_DATA_FOLDER) || mkdir -p $(RAW_DATA_FOLDER)
+	wget $(URIBASE)/$*.owl -O $@
+
+$(RAW_DATA_FOLDER)/mp.owl:
+	test -d $(RAW_DATA_FOLDER) || mkdir -p $(RAW_DATA_FOLDER)
+	wget $(URIBASE)/mp/mp-base.owl -O $@
+
+$(RAW_DATA_FOLDER)/hp-mp-merged.owl: $(RAW_DATA_FOLDER)/hp.owl $(RAW_DATA_FOLDER)/mp.owl
+	test -d $(RAW_DATA_FOLDER) || mkdir -p $(RAW_DATA_FOLDER)
+	robot merge --input $(RAW_DATA_FOLDER)/hp.owl --input $(RAW_DATA_FOLDER)/mp.owl reason reduce --output $@
```
matentzn, line 34: Please create new issue: `add imports module to hp-mp-merged.owl`

Makefile
```diff
@@ -0,0 +1,74 @@
...
+
+
+
+$(RAW_DATA_FOLDER)/random-hp-terms.txt:
+	$(eval TERM=$(shell bash -c "cat $(RAW_DATA_FOLDER)/hp-mp-merged.owl | grep '<!-- http://purl.obolibrary.org/obo/HP_'  | grep '_' |  shuf -n 100  | head -n 100 | rev | cut -d '/' -f 1 | cut -d ' ' -f 2 | rev | tr '_' ':'"))
+	echo $(TERM) > $@
+
+$(RAW_DATA_FOLDER)/random-mp-terms.txt:
+	#TODO: GET BACK TO HP
+	$(eval TERM=$(shell bash -c "cat $(RAW_DATA_FOLDER)/hp-mp-merged.owl | grep '<!-- http://purl.obolibrary.org/obo/HP_'  | grep '_' |  shuf -n 100  | head -n 100 | rev | cut -d '/' -f 1 | cut -d ' ' -f 2 | rev | tr '_' ':'"))
```
matentzn, line 54:
```suggestion
	$(eval TERM=$(shell bash -c "cat $(RAW_DATA_FOLDER)/hp-mp-merged.owl | grep '<!-- http://purl.obolibrary.org/obo/MP_'  | grep '_' |  shuf -n 100  | head -n 100 | rev | cut -d '/' -f 1 | cut -d ' ' -f 2 | rev | tr '_' ':'"))
```

tox.ini
```diff
@@ -49,6 +49,7 @@ description = Run the flake8 tool with several plugins (bandit, docstrings, impo
 #########################
 [flake8]
 ignore =
+    BLK100
```
matentzn, line 52: Not a fan. Why?

Makefile
matentzn: Still does not seem addressed?

matentzn (commented):
One open comment left, then you can merge, no need to wait for approval

Makefile
souzadevinicius: Sorry!

matentzn (approved):

matentzn:
Very nice work @souzadevinicius!

PR #54, opened 2023-02-06
#53 #55  post-processing methods - creation of pheval output

yaseminbridges:
These methods include sorting tool-specific output that has been post-processed to the PhEval gene/variant dataclasses in descending order with "best" scores first and assigning ranks to these results.

### DIFF
--- a/src/pheval/post_processing/post_processing.py
+++ b/src/pheval/post_processing/post_processing.py
@@ -62,7 +62,7 @@ class RankedPhEvalVariantResult:
 
 class ResultSorter:
     def __init__(
-            self, pheval_results: [PhEvalGeneResult] or [PhEvalVariantResult], ranking_method: str
+        self, pheval_results: [PhEvalGeneResult] or [PhEvalVariantResult], ranking_method: str
     ):
         self.pheval_results = pheval_results
         self.ranking_method = ranking_method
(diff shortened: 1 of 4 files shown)


yaseminbridges:
> Looks great. Remember, when I approve something, I didn't do an in-depth review, only an architectural one, so you should get a proper review from @souzadevinicius and/or @julesjacobsen if you want to be sure.

No problem, I've requested both of their reviews for this as well

yaseminbridges:
Just added 2 small methods to write the pheval gene and variant .tsv results

src/pheval/post_processing/post_processing.py
```diff
@@ -57,3 +61,103 @@ def as_dict(self):
...
+        """Sort results with best score first."""
+        return (
+            self.sort_by_increasing_score()
+            if self.ranking_method.lower() == "pvalue"
+            else self.sort_by_decreasing_score()
+        )
+
+
+@dataclass
+class ScoreRanker:
```
julesjacobsen, line 91: add enum for ascending/descending ranking order. e.g. p-value is in ascending order 0.0 = rank 1, combined score is in descending order 1.0 = rank 1.

The score and sort/rank order should be defined in the config file - add in an explanation otherwise the scores might be ranked in the wrong order!

julesjacobsen (changes requested):
Implement the sort order checks and input in the config file

src/pheval/post_processing/post_processing.py
yaseminbridges: tried to add in the match case statements we spoke about yesterday for throwing errors when results were not correctly sorted - however I think this is something implemented from python 3.10 onwards and so it failed when running `tox`. So instead I am using if/elif for the time being

julesjacobsen (approved):

PR #65, opened 2023-03-03
56 merge multi input and single input commands

### LINKED ISSUE #56: Is there a reason we have separate commands for single input and multi input?
It seems odd to have separate commands every time we want to be able to pass a single element vs a set of elements. Usually, it is better to solve this on option level.

So rather than:

`pheval-utils scramble-phenopacket` / `pheval-utils scramble-phenopackets` 

this is what I usually see:

`pheval-utils scramble-phenopacket --phenopacket filename.json` / `pheval-utils scramble-phenopacket --phenopackets corpus.json` (or directory, doesn't matter).

@yaseminbridges can you explain your design here? Maybe I am wrong about that! :)

yaseminbridges:
Closes #56 

Merged multi-input and single-input commands into a single command for:
- Updating the gene identifiers and symbols in phenopackets
- Scrambling the phenotypic profiles in phenopackets
- Creating spiked vcfs

Also addressed some prior points of moving the click commands to the cli_pheval_utils.py and keeping it completely separate to the methods, as well as removing the "hacking" of file names in the scramble phenopackets command

### DIFF
--- a/src/pheval/cli.py
+++ b/src/pheval/cli.py
@@ -7,12 +7,12 @@ from pheval.analyse.analysis import benchmark, benchmark_comparison
 
 from .cli_pheval import run
 from .cli_pheval_utils import (
+    create_spiked_vcfs_command,
     scramble_phenopackets_command,
     scramble_semsim,
     semsim_comparison,
     update_phenopackets_command,
 )
-from .prepare.create_spiked_vcf import create_spiked_vcf_command, create_spiked_vcfs_command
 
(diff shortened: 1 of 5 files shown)


matentzn (approved):
I very much agree with the general design decisions, excellent! I have only done a cursory "architectural" review, not looking at any specific python code (mostly CLI).

Great work, thanks!

julesjacobsen (approved):

PR #70, opened 2023-03-08
66 extend pheval runner with getters for subdirectory structure

yaseminbridges:
Implemented getters for the various output subdirectories to be created during a run, as well as a function that builds the output directory layout prior to the ```prepare``` step in the run command. Eliminating the need for developers to create the directory layout in the runner implementation


yaseminbridges:
Expanded the PhEvalRunner class with getters for the various output subdirectories specified in the slides for a full `pheval run`. I have defined set properties within the abstract PhEvalRunner class that can be called from runners. For example, within the Exomiser Runner, if you wanted to retrieve the directory for the tool specific results, you can use ```self.runner_results_dir``` within the runner class.

For reference, given an output directory of ```/results``` in the CLI:

```self.runner_version_dir``` --> ```/results/exomiser-13.2.0```
```self.runner_input_commands_dir``` --> ```/results/exomiser-13.2.0/runner_input_commands```
```self.corpus_variant_dir``` --> ```/results/exomiser-13.2.0/default-corpus1-default```
```self.runner_results_dir``` --> ```/results/exomiser-13.2.0/default-corpus1-default/results```
```self.pheval_gene_results_dir``` --> ```/results/exomiser-13.2.0/default-corpus1-default/pheval_gene_results```
```self.pheval_variant_results_dir ``` --> ```/results/exomiser-13.2.0/default-corpus1-default/pheval_variant_results```

I have also implemented a concrete method within the PhEvalRunner abstract class that will create the output subdirectories prior to the `prepare`, `run`, and `post-process` steps - to prevent the creation of paths in a runner implementation. I'm using a very basic idea of a input dir config to get this all running and have created tests for the getters/setters and creation of output sub directories. 

@matentzn is this what you was looking for?

File: src/pheval/config_parser.py [hunk 1]
```diff
@@ -0,0 +1,24 @@
+from dataclasses import dataclass
+from pathlib import Path
+
+import yaml
+from serde import serde
+from serde.yaml import from_yaml
+
+
+@serde
```
matentzn, line 9: What does this do?

src/pheval/config_parser.py
```diff
@@ -0,0 +1,24 @@
...
+@dataclass
+class InputDirConfig:
```
matentzn, line 11: Can you add some documentation what exactly this class is for?

src/pheval/config_parser.py
```diff
@@ -0,0 +1,24 @@
...
+    tool: str
+    software_directory: Path or None
+    software_config: Path or None
+    input_directory: Path
+    phenotype_only: bool
```
matentzn, line 16: This seems out of place here. Can we add some documentation what each of these parameters exactly are?

matentzn: (maybe this is not out of place, just need to understand why it is here)

src/pheval/runners/runner.py
```diff
@@ -14,6 +22,91 @@ class PhEvalRunner(ABC):
     output_dir: Path
     config_file: Path
     version: str
+    directory_path = None
+
+    def _get_tool(self):
+        return parse_input_dir_config(self.input_dir).tool
```
matentzn, line 28: From the parse_input_dir_config method, it seems the parameter it expects the path to a config file, not a directory.

src/pheval/runners/runner.py
```diff
@@ -14,6 +22,91 @@ class PhEvalRunner(ABC):
...
+
+    def _get_phenotype_only(self):
+        return parse_input_dir_config(self.input_dir).phenotype_only
```
matentzn, line 31: From the parse_input_dir_config method, it seems the parameter it expects the path to a config file, not a directory.

src/pheval/config_parser.py
```diff
@@ -0,0 +1,24 @@
...
+import yaml
+from serde import serde
+from serde.yaml import from_yaml
+
+
+@serde
+@dataclass
+class InputDirConfig:
+    tool: str
+    software_directory: Path or None
```
matentzn, line 13: Is this possibley redundant? Isnt this the same as "the directory the software_config is in"?

src/pheval/runners/runner.py
matentzn: Also, I think its ok to read this file over and over again, but maybe just reading it once would be a better design. Up to you.

src/pheval/utils/constants.py
```diff
@@ -0,0 +1,4 @@
+PHEVAL_GENE_RESULTS_DIR = "pheval_gene_results/"
+PHEVAL_VARIANT_RESULTS_DIR = "pheval_variant_results/"
+TOOL_RESULTS_DIR = "results/"
+RUNNER_INPUT_COMMANDS_DIR = "runner_input_commands/"
```
matentzn, line 4: remind me why this is here? What was this again?

matentzn (commented):
Great job - I have a few questions, but probably nothing serious!

src/pheval/config_parser.py
yaseminbridges: This has been the library I'm using to parse any yaml config (I started using it in the Exomiser runner). So it will parse the contents into the dataclass so I would be able to access elements like so: ```input_dir_config.tool```, while before I would be accessing it as a dictionary element ```input_dir_config["tool"]```.

It will work if there are extra fields in the config that are not documented in the dataclass but I'm not exactly sure if these can be accessed if not specified in the dataclass

src/pheval/runners/runner.py
yaseminbridges: Yes, it expects the config file for the input dir, I thought this would be the input now for the `--input-dir` parameter? If not I can set it back to the directory and specify the file name from there to read

src/pheval/runners/runner.py
yaseminbridges: I will alter to just reading once

src/pheval/config_parser.py
yaseminbridges: Yes, it could be redundant, I just wasn't sure how the input directory would be laid out, whether for example the application.properties for Exomiser would be put just in the input directory:

```
└── exomiser-13.2.0
    └── default
        ├── 2302_hg19
        ├── 2302_phenotype
        ├── application.properties
        └── exomiser-cli-13.2.0
```

or within the exomiser-cli directory

```

└── exomiser-13.2.0
    └── default
        ├── 2302_hg19
        ├── 2302_phenotype
        └── exomiser-cli-13.2.0
            └── application.properties
```

src/pheval/runners/runner.py
yaseminbridges: Ah wait yes, I think I misunderstood from last week, a directory is still expected as the parameter for the `--input-dir` but as a developer, you can also expect a `config.yaml` within that directory that you can access?

src/pheval/config_parser.py
matentzn: Cool!

matentzn: cc @souzadevinicius just FYI, this is something interesting I just learned and never used before!

src/pheval/runners/runner.py
matentzn: > Ah wait yes, I think I misunderstood from last week, a directory is still expected as the parameter for the --input-dir but as a developer, you can also expect a config.yaml within that directory that you can access?

Exactly!

src/pheval/config_parser.py
matentzn: What is the advantage to give structure, rather then simply a flat list (first example)?

src/pheval/utils/constants.py
matentzn: No I meant, what is this DIR: `RUNNER_INPUT_COMMANDS_DIR`?

src/pheval/runners/runner.py
yaseminbridges: Understood, will make those changes back to expecting a directory

src/pheval/config_parser.py
yaseminbridges: Sure, it does give it structure, if the location of the software config was in the software directory, it makes sense, and as you said it can allow for one less field in the config. Whatever you feel works better structure wise I'm happy to go with and make the changes

src/pheval/utils/constants.py
yaseminbridges: This is the tool input commands directory, i.e. for Exomiser, depending on whether you want to run the jobs in batches or all in one go, the directory will contain the file(s) (which contains the commands) for exomiser to run the jobs. 

`--sample /Users/yaseminbridges/full_test_pheval_run/corpora/corpus1/default/phenopackets/Abdul_Wahab-2016-GCDH-Patient_5.json --output-directory /Users/yaseminbridges/full_test_pheval_run/output/exomiser-13.2.0/default-corpus1-default/results --output-filename Abdul_Wahab-2016-GCDH-Patient_5-exomiser --preset phenotype-only`

matentzn: Why should this be fixed? Isnt this passed in as a command line parameter?

yaseminbridges: Not within the PhEval run command. It's this path `/results/exomiser-13.2.0/runner_input_commands` -  during `pheval run` in the `run` step the batch files are prepared and output to this directory, and then exomiser/phen2gene runs using that batch file.

The naming for the 'batch' file needs to change to the same naming as {config}-{corpus}-{corpus-variant}-batch.txt, but this is the idea. Jules said it would be a good idea to keep the actual tool commands used to produce this output in case whoever is running it wants to run the jobs again but separate to the whole `pheval run`
```
├── exomiser-13.2.0
│   ├── default-corpus1-default
│   │   ├── pheval_gene_results
│   │   │   ├── Abdul_Wahab-2016-GCDH-Patient_5-pheval_gene_result.tsv
│   │   │   └── Ajmal-2013-BBS1-IV-5_family_A-pheval_gene_result.tsv
│   │   └── results
│   │       ├── Abdul_Wahab-2016-GCDH-Patient_5-exomiser.html
│   │       ├── Abdul_Wahab-2016-GCDH-Patient_5-exomiser.json
│   │       ├── Ajmal-2013-BBS1-IV-5_family_A-exomiser.html
│   │       └── Ajmal-2013-BBS1-IV-5_family_A-exomiser.json
│   └── runner_input_commands
│       └── default-exomiser-batch.txt
```

The directory `runner_input_commands` probably needs a better name so it can be separated from the `runner`, maybe `tool_input_commands`/`tool_commands` make more sense?

src/pheval/config_parser.py
yaseminbridges: added some documentation following how @souzadevinicius has structured his docstrings

src/pheval/runners/runner.py
yaseminbridges: added a __post_init__ method that reads the input config once

src/pheval/utils/constants.py
yaseminbridges: renamed directory to `tool_input_commands` to reduce confusion with the runner

src/pheval/config_parser.py
yaseminbridges: removed this field within the input dir config

src/pheval/config_parser.py
yaseminbridges: added some documentation following how @souzadevinicius has structured his docstrings

src/pheval/runners/runner.py
```diff
@@ -14,6 +22,95 @@ class PhEvalRunner(ABC):
     output_dir: Path
     config_file: Path
     version: str
+    directory_path = None
+    input_dir_config = None
+
+    def __post_init__(self):
```
matentzn, line 28: I did not know this was possible. Cool, learned something new.

src/pheval/runners/runner.py
```diff
@@ -14,6 +22,95 @@ class PhEvalRunner(ABC):
...
+        return Path(self.output_dir).joinpath(f"{self._get_tool()}-{self.version}")
+
+    @runner_version_dir.setter
+    def runner_version_dir(self, directory_path):
+        self.directory_path = Path(directory_path)
+
+    @property
+    def runner_input_commands_dir(self):
+        return Path(self.output_dir).joinpath(
+            f"{self._get_tool()}-{self.version}/{RUNNER_INPUT_COMMANDS_DIR}"
```
matentzn, line 48: This is the last thing I am not sure about. Why isn't this simply the `--input-dir` value?

matentzn: I mean, the directory path is formatted correctly, but I thought the logic of the path creation should reside in the `Makefile`, not here..

matentzn: Is this conceptually like a temporary directory where the runner may choose to deposit intermediate files?

yaseminbridges: I think it is something that can be put in the temporary directory, it's not an important runner output in terms of the results you want to get out of PhEval. It's just something that can allow you to run the exact same samples with the settings outside of PhEval. Normally I use the `prepare_batch_files` alone when I benchmark Exomiser on different environments. I can add a path to the `temp_dir` that has this directory, and then if the `--temp-dir` is specified in the CLI then the path is created there, otherwise in the output directory? What do you suggest?

yaseminbridges: It's not so much to do with the input dir, more so with the test data dir. I also don't think it is a step that can be skipped as I don't know any tools that can take a phenopacket directory or vcf directory alone and produce the results from there.

For Phen2Gene, to run on the command line you would need to run this (outside of PhEval)
```
python3 /Users/yaseminbridges/phen2gene/Phen2Gene/phen2gene.py --manual HP:0000256 HP:0002059 HP:0100309 HP:0003150 HP:0001332 -out ../../../full_test_pheval_run/output/phen2gene/corpus1_results/phen2gene_results/ --name Abdul_Wahab-2016-GCDH-Patient_5 -d /Users/yaseminbridges/phen2gene/Phen2Gene/lib
```

In the runner implementation there is a single file that contains a command like this for all the phenopacket samples, i.e. there are 385 lines within the file that contain all the commands for running the corpus (385 samples within that corpus). So in the PhEval run, you create this file that has all the commands and then this is run `bash {path/to/tool_input_commands_dir/file.txt}`

So it is using a mixture of data from the input dir and testdata dir but I'm not sure if it's really preprocessing any data - it's just the actual commands needed to run the tool.

matentzn: > I think the idea I had as well is that there could be fields in the config.yaml within the tool_specific_configurations that could allow for I guess a more specific running of a software, i.e. extra parameters that can be used could be specified here if they do not always want to use it as a default

For sure.

>  I had just gone with the idea that anything output by pheval run goes in the output dir.

You are right, ok. Thanks for persisting and changing my mind. Since an output dir is already connected to a specific tool/version/corpus combination, you should be able to simply call it 

```
        return Path(self.output_dir).joinpath(f"{TOOL_INPUT_COMMANDS_DIR}"
```

Right? Or is there a point of adding `{self._get_tool()}-{self.version}/` in there?

yaseminbridges: For sure, there are two different ways of going about it, I can either give it a general directory within the `output_dir`

```python
return Path(self.output_dir).joinpath(f"{TOOL_INPUT_COMMANDS_DIR}")
```
 and then the name of the file/prefix will need to be specific to the run, i.e. `exomiser-13.2.0-default-corpus1-default.txt` so that the `run` command can precisely locate the correct file containing the commands and only run the ones specific to that single run, to avoid running previous runs. This prefix is something that I can provide in the getters.

Alternatively, you can give it a specific directory:

```python
return Path(self.output_dir).joinpath(f"{self._get_tool()}-{self.version}/{TOOL_INPUT_COMMANDS_DIR}")
```

and a less specific file name `default-corpus1-default.txt`.

Or even more specific directory location...

Whichever you prefer I can do, the file just needs to be named something that can be identifiable by the runner to only run those commands and not previously generated files from other runs

matentzn: In the way I thought about it, `Path(self.output_dir)` should be a directory which stores the results of a _specific run_. So it already is clear which exact corpus, variant / version of tool etc was used (results.yml). Why would you need to name the output directory at all?

matentzn: That was my understanding, exactly!

yaseminbridges: I had no idea! I can remove some of the getters that I have then as they are redundant. So the building of the output directory structure:

```
output
├── exomiser-13.2.0
│   ├── default-corpus1-default
│   │   ├── pheval_gene_results
│   │   ├── pheval_variant_results
│   │   └── results
│   └── runner_input_commands
```
 is going to be handled by the MakeFile? Or do I still need to build it before the prepare step in pheval run?

matentzn: The way I was thinking of it, the Makefile will create:

```
output
├── exomiser-13.2.0
│   ├── default-corpus1-default
```

And pheval will create will create the rest!

```
pheval_gene_results
 pheval_variant_results
 results
```

Continue to challenge me! I trust you more than myself :P

yaseminbridges: No problem, thank you with bearing with me too!

src/pheval/runners/runner.py
```diff
@@ -3,6 +3,14 @@
 from dataclasses import dataclass
 from pathlib import Path
 
+from pheval.config_parser import parse_input_dir_config
+from pheval.utils.constants import (
+    PHEVAL_GENE_RESULTS_DIR,
+    PHEVAL_VARIANT_RESULTS_DIR,
+    TOOL_INPUT_COMMANDS_DIR,
+    TOOL_RESULTS_DIR,
+)
```
matentzn, line 12: I guess Jules will tell you to add these as member vars into the class, but good luck.

matentzn (approved):
Very nice! I am sure your constants will be moved by Jules,but I am happy the way it is.

src/pheval/runners/runner.py
julesjacobsen: For real! These aren't used anywhere else and only have context inside that class, so why externalise them?

src/pheval/runners/runner.py
yaseminbridges: discussed with Jules and he is happy with the PR as is, but we have suggested some changes on the naming of the output directories as well as the paths/information that the pheval analysis command needs to know, see #90

PR #71, opened 2023-03-08
67 add version option to pheval run

yaseminbridges:
Adding ```version``` option to pheval run command

### DIFF
--- a/src/pheval/cli_pheval.py
+++ b/src/pheval/cli_pheval.py
@@ -1,6 +1,7 @@
 """
 Monarch Initiative
 """
+from pathlib import Path
 
 import click
 
@@ -14,6 +15,7 @@ from pheval.implementations import get_implementation_resolver
     metavar="INPUTDIR",
     required=True,
     help="The input directory (relative path: e.g exomiser-13.11)",
+    type=Path,
 )
 @click.option(
     "--testdata-dir",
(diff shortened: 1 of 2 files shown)


src/pheval/cli_pheval.py
```diff
@@ -35,13 +38,15 @@
...
 @click.option(
     "--output-dir",
     "-o",
     metavar="OUTPUTDIR",
     required=True,
     help="The path of the output directory",
+    type=Path,
 )
 @click.option(
     "--config",
```
matentzn, line 52: I know this is not what the PR is about, but have you thought about my suggestion to ditch this?

matentzn (approved):
Excellent @yaseminbridges thank you!

src/pheval/cli_pheval.py
yaseminbridges: Ditching the ```--config``` option? I think I thought about it too much and confused myself with what you meant. Do you mean that the config attached with the ```--input-dir``` parameter would have the information contained within the config attached to the ```--config-file``` ? 

I don't know if it could be 'fixed' in terms of it's contents as the different implementations require different levels of information to run, for example, LIRICAL needs both a path to its own database and the path to the Exomiser variants database for it to output variant prioritisation results (but I guess this information would be attached to the ```--input-dir``` config anyway?). If the MakeFile is dealing with the retrieval of the databases for different implementations and can write these paths to the ```--input-dir``` config automatically then I think it's doable

matentzn: Lets talk about this tomorrow, can you add to agenda? Nothing to do with this PR here. No need to wait

yaseminbridges: No problem!

PR #72, opened 2023-03-08
69 update phenopackets corpus

yaseminbridges:
Updated the phenopackets corpus:

- Removed phenopackets that were duplicated
-  Added missing phenopackets
-  Removed non latin letters from filenames

There are now 385 phenopackets within the corpus

(No CLI changes just updated the phenopackets)


matentzn (approved):
Thanks for making a PR for this!

PR #83, opened 2023-03-16
80 obtain negated phenotypic features from a phenopacket

yaseminbridges:
Added a method within the `PhenopacketUtil` class to obtain any phenotypic features marked as excluded. Also added tests to test its functionality. No CLI changes

### DIFF
--- a/src/pheval/utils/phenopacket_utils.py
+++ b/src/pheval/utils/phenopacket_utils.py
@@ -109,7 +109,7 @@ class PhenopacketUtil:
             phenotypic_features.append(p)
         return phenotypic_features
 
-    def obtain_negated_phenotypes(self) -> [PhenotypicFeature] or None:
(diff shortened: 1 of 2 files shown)


matentzn (approved):
LGTM

src/pheval/utils/phenopacket_utils.py
```diff
@@ -109,6 +109,15 @@ def observed_phenotypic_features(self) -> list[PhenotypicFeature]:
             phenotypic_features.append(p)
         return phenotypic_features
 
+    def negated_phenotypic_features(self) -> [PhenotypicFeature] or None:
```
julesjacobsen, line 112: Is this a friendly thing to do in Python? This would hurt in Java where an empty list is preferred.

yaseminbridges: Changed it!

PR #84, opened 2023-03-16
81 obtain the sample id from a phenopacket

yaseminbridges:
Added method to the `PhenopacketUtil` class to retrieve the sample id for the proband. Also added tests for its functionality. No CLI changes

### DIFF
--- a/src/pheval/utils/phenopacket_utils.py
+++ b/src/pheval/utils/phenopacket_utils.py
@@ -92,6 +92,13 @@ class PhenopacketUtil:
     def __init__(self, phenopacket_contents: Phenopacket):
         self.phenopacket_contents = phenopacket_contents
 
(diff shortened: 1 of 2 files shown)


matentzn (approved):

src/pheval/utils/phenopacket_utils.py
```diff
@@ -92,6 +92,13 @@ class PhenopacketUtil:
     def __init__(self, phenopacket_contents: Phenopacket):
         self.phenopacket_contents = phenopacket_contents
 
+    def sample_id(self) -> str:
+        """Retrieve the sample ID."""
```
julesjacobsen, line 96: Change the doc string to say """Retrieve the sample ID from a phenopacket or proband of a family."""

PR #85, opened 2023-03-16
82 calculate end pos for variant from start and ref allele

yaseminbridges:
Added method to the `post_processing.py` to calculate the end position for a variant given the start position and the reference allele. This is needed as some VP tools do not return the end position of a variant, this method will need to be called by the implementor so that the results can be standardised to the pheval variant format. Also added tests for its functionality. No CLI changes

### DIFF
--- a/src/pheval/post_processing/post_processing.py
+++ b/src/pheval/post_processing/post_processing.py
@@ -6,6 +6,11 @@ from pathlib import Path
 import pandas as pd
 
 
+def calculate_end_pos(variant_start: int, variant_ref: str) -> int:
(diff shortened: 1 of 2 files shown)


matentzn (approved):
These PRs are what I am talking about. Remeber when I do reviews, its only high level, I don't think about the specific methods!

yaseminbridges:
> These PRs are what I am talking about. Remeber when I do reviews, its only high level, I don't think about the specific methods!

Will aim for PRs of this size! I understand, I usually go into a deeper review of the methods with Jules

PR #86, opened 2023-03-16
79 obtain the gene symbol from a gene identifier

yaseminbridges:
Added methods to `GeneIdentifierUpdater` class to retrieve the gene symbol from a gene identifier. Also added tests for its functionality. No CLI changes

### DIFF
--- a/src/pheval/utils/constants.py
+++ b/src/pheval/utils/constants.py
@@ -1,0 +1,4 @@
+ENSEMBL_ID = "ensembl_id"
+HGNC_ID = "hgnc_id"
+ENTREZ_ID = "entrez_id"
+REFSEQ_ACCESSION = "refseq_accession"
--- a/src/pheval/utils/constants.py
+++ b/src/pheval/utils/constants.py
@@ -1,4 +1,0 @@
-ENSEMBL_ID = "ensembl_id"
-HGNC_ID = "hgnc_id"
-ENTREZ_ID = "entrez_id"
-REFSEQ_ACCESSION = "refseq_accession"
--- a/tests/test_phenopacket_utils.py
+++ b/tests/test_phenopacket_utils.py
@@ -21,7 +21,7 @@ from phenopackets import (
 )
 
 from pheval.prepare.custom_exceptions import IncorrectFileFormatError
-from pheval.utils.constants import HGNC_ID, ENTREZ_ID
+from pheval.utils.constants import ENTREZ_ID, HGNC_ID
 from pheval.utils.phenopacket_utils import (  # GeneIdentifierUpdater,; create_hgnc_dict,
     GeneIdentifierUpdater,
     GenomicVariant,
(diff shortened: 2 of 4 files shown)


tests/test_phenopacket_utils.py
```diff
@@ -480,6 +480,22 @@ def test_find_identifier(self):
         self.assertEqual(self.gene_identifier_updater_entrez.find_identifier("A2M"), "2")
         self.assertEqual(self.gene_identifier_updater_entrez.find_identifier("GBA"), "2629")
 
+    def test_obtain_gene_symbol_from_identifier_hgnc(self):
+        self.assertEqual(
+            self.gene_identifier_updater_ens.obtain_gene_symbol_from_identifier(
+                "HGNC:5", "hgnc_id"
```
matentzn, line 486: For things like keys in standard dicts, I recommend using constants, because of the likely of change later down the line

HGNC_ID = "hgnc_id" in constants.py

and here:

```suggestion
                "HGNC:5", HGNC_ID
```

matentzn: But feel free to ignore if it blocks your flow or you disagree.

yaseminbridges: I agree, didn't think of that!

yaseminbridges:
added the constants for the dictionary, referencing these instead of strings throughout the code

matentzn (approved):
Excellent @yaseminbridges

src/pheval/utils/constants.py
julesjacobsen, line 1: I think these would be better not being here at all - this is all implementation details which should be hidden in the class, not shared around.

A cleaner (but less performant) API would be simply:

```python

def obtain_gene_symbol_from_identifier(self, query_gene_identifier: str) -> str:
    # lookup pattern e.g. ENSG, ncbigene: , number, HGNC: ...
   # create a dictionary identifier: symbol 
    return symbol from identifiers dict
```

julesjacobsen: You don't want to share these constants, it's too much information and leaking implementation details.

matentzn: Ok, I was not aware this is bad practice. My main point is to never use literal strings as keys for retrieving information from dictionaries that _are_ shared around, because it violates the single responsibility principle (lets say you want to change a key from "hgnc_id" to  "hgnc", you will have to find all places in the code where "hgnc_id" is used

julesjacobsen: This is what we want:

User: "Hey API, what's the gene symbol for ENSG00012334?"
API "ENSG00012334 is GENE1"

not 

User "Hey API, what's the gene symbol for ENSG00012334 using the key type of ENSEMBL_ID"
API "ENSEMBL_ID: ENSG00012334 is GENE1"

Why not? Because ENSG00012334 , 123456, HGNC:12345 are all natural keys for starters, secondly, what if you asked for ENSG00012334 with the NCBI_ID key? Thirdly, why are you even exposing the fact that the API uses a map to store this info? None of it is required it's simply a bi-directional multi-map internally, externally you just want a simple request to get the gene symbol for a limited set of gene identifiers.

matentzn: Ahhhh right! Yes, that makes total sense now!

julesjacobsen (approved):

PR #88, opened 2023-03-21
adding 281 samples from phen2gene paper

yaseminbridges:
I have added 281 phenopackets, converted from the Phen2Gene testing data to the testdata directory. 

These phenopackets only contain the solved disease-causing genes. So it is important to note that you can not spike vcfs with these phenopackets and they can only be run on tools that can run on only the HPO term data only (gene prioritisation), i.e. Phen2Gene or Exomiser once [PR #3](https://github.com/monarch-initiative/pheval.exomiser/pull/8) is completed for pheval.exomiser


julesjacobsen:
I think the genomic interpretation call in each Phenopacket needs to be a GeneDescriptor, not a VariantInterpretation as the VariantInterpretation requires a variant (due to multiplicity 1..1), and this information is missing.

See

https://phenopacket-schema.readthedocs.io/en/latest/genomic-interpretation.html

https://phenopacket-schema.readthedocs.io/en/latest/gene.html#rstgene

yaseminbridges:
> I think the genomic interpretation call in each Phenopacket needs to be a GeneDescriptor, not a VariantInterpretation as the VariantInterpretation requires a variant (due to multiplicity 1..1), and this information is missing.
> 
> See
> 
> https://phenopacket-schema.readthedocs.io/en/latest/genomic-interpretation.html
> 
> https://phenopacket-schema.readthedocs.io/en/latest/gene.html#rstgene

Yes! I missed this, will make the changes to the phenopackets

yaseminbridges:
changed the genomic interpretation call in each Phenopacket to a GeneDescriptor

julesjacobsen:
Are there no diagnosed conditions for these cases?

yaseminbridges:
> Are there no diagnosed conditions for these cases?

No, I only have the disease-causative genes and the HPO ids for these cases

PR #93, opened 2023-03-28
adding structural variant phenopackets

yaseminbridges:
182 structural phenopackets added to the corpora, these have been converted to the v2 phenopackets schema. It should be noted that the variants in these phenopackets are based on the GRCh38 human genome build and so shouldn't be spiked in the template exome VCF currently provided in PhEval. Instead they need to be spiked in a benchmarking structural variants VCF for the correct build. Damian previously used the ones made available from Genome in a bottle from this trio: https://ftp-trace.ncbi.nlm.nih.gov/ReferenceSamples/giab/release/AshkenazimTrio/, I'm not sure if he used a specific sample but probably any would be fine. These are too large to upload onto the GitHub, so not sure how to overcome this apart from specifying where it could be downloaded from at the moment


testdata/phenopackets/structural_variants/Agrawal-2014-SPEG-P2.json
```diff
@@ -195,7 +203,8 @@
         "name": "NCBI organismal classification",
         "url": "http://purl.obolibrary.org/obo/ncbitaxon.owl",
         "version": "2018-03-02",
-        "namespacePrefix": "NCBITaxon"
+        "namespacePrefix": "NCBITaxon",
+        "iriPrefix": "http://purl.obolibrary.org/obo/NCBITAXON_"
```
matentzn, line 207:
```suggestion
        "iriPrefix": "http://purl.obolibrary.org/obo/NCBITaxon_"
```

yaseminbridges:
These phenopackets all now pass with the `validate` in `phenopacket-tools`.

julesjacobsen:
VCF spec - https://samtools.github.io/hts-specs/VCFv4.3.pdf

yaseminbridges:
> Ensure that any symbolic variants have a corresponding `INFO` field to describe the end position, type and length.
> 
> VCF spec - https://samtools.github.io/hts-specs/VCFv4.3.pdf
> 
> FORMAT should be like this:
> 
> ```
> SVTYPE=INS;END=12345;SVLEN=200;
> SVTYPE=DEL;END=12345;SVLEN=-200;
> ```

submitted issue to [phenopacket-tools](https://github.com/phenopackets/phenopacket-tools/issues/184) which should help with including the `info` in the `VcfRecord` when converting from v1 to v2

yaseminbridges:
Need to fix the `SVLENGTH` AND `END` in the info fields

PR #94, opened 2023-03-29
74 provide provisional output for existing runners

yaseminbridges:
Extended the PhEval Runner with a method to create the metadata for a run. By default, with no alteration by any runner implementations, the results.yml would output something like so:

```
tool: exomiser
tool_version: 13.2.0
config: exomiser-13.2.0-default
run_timestamp: 1680092494
corpus: corpus1-default
tool_specific_configuration_options: Null
```

The tool specific configurations can be overwritten by the tool developer in the runner class like so:
```python
def construct_meta_data(self):
     meta_data = self.meta_data
     meta_data.tool_specific_configuration_options = ExomiserToolConfigurations(hg19_version='2209',
                                                                                hg38_version='2209',
                                                                                phenotype_version='2302')
     return metadata
```

Alternatively, handing the metadata already created to another method separate to the runner:

```python
def construct_meta_data(self):
     return write_tool_specific_configs(self.meta_data) # where the metadata data class is being returned in the method
```

Ideally, the implementor would assign tool_specific_configuration_options a dataclass, which can be as nested/detailed as required and the field will be overwritten. The above method would return this:

```
tool: exomiser
tool_version: 13.2.0
config: exomiser-13.2.0-default
run_timestamp: 23
corpus: corpus1-default
tool_specific_configuration_options:
  hg19_version: '2209'
  hg38_version: '2209'
  phenotype_version: '2302'
```

### DIFF
--- a/src/pheval/runners/runner.py
+++ b/src/pheval/runners/runner.py
@@ -1,9 +1,8 @@
 """Runners Module"""
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from datetime import datetime
 from pathlib import Path
-
-from google.protobuf.timestamp_pb2 import Timestamp
 
 from pheval.config_parser import parse_input_dir_config
 from pheval.run_metadata import BasicOutputRunMetaData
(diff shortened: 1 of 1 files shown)


src/pheval/runners/runner.py
```diff
@@ -3,7 +3,10 @@
 from dataclasses import dataclass
 from pathlib import Path
 
+from google.protobuf.timestamp_pb2 import Timestamp
```
matentzn, line 6: Thats an odd dependency to have in pheval.. Do we really need this?

matentzn (approved):
Looks great! Just one comment, but LGTM

src/pheval/runners/runner.py
yaseminbridges: Yes, you're right! I can use a built-in module for this

PR #102, opened 2023-04-12
removing files object from phenopackets

yaseminbridges:
Instead of changing the paths to the VCF files in the phenopackets I have removed the `files` object entirely from the phenopackets that had them.


matentzn:
I cant speak to this - this is best reviewed carefully by Jules as it appears at first glance to remove some important metadata; so better not rely on my intuition here.

julesjacobsen:
This is fine as the files names and properties are fictitious and made up on-the-fly later on so they are safe to delete. All tools (so far, at least) allow specifying the VCF independently on the CLI.

PR #103, opened 2023-04-13
100 merge post processing methods

yaseminbridges:
Added an additional method `generate_pheval_result()` to be called by the implementor for the final post-processing steps. This method will rank the results and format them into the tsv results depending on whether the implementor provides it with a list of variant or gene results.

Changed public methods to private methods (in the pythonic way) to hopefully document that these should not be called in a runner implementation.


matentzn (approved):
Looks great!

julesjacobsen (approved):
Very nice.

PR #107, opened 2023-04-17
106 formatting time at last encounter field

yaseminbridges:
Correctly formatting the `time_at_last_encounter` field in the phenopacket `subject`. This included correctly formatting to iso8601duration, removing n/a values and adding `ontologyClass` fields where necessary to describe the age.

### DIFF
--- a/testdata/phenopackets/lirical/Abdul_Wahab-2016-GCDH-Patient_5.json
+++ b/testdata/phenopackets/lirical/Abdul_Wahab-2016-GCDH-Patient_5.json
@@ -2,7 +2,6 @@
   "id": "PMID:27672653-Abdul_Wahab-2016-GCDH-Patient_5",
   "subject": {
     "id": "Patient 5",
-    "dateOfBirth": "1970-01-01T00:00:00Z",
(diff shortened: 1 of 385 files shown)


julesjacobsen:
This looks good. Only one comment, every dateOfBirth is `"dateOfBirth": "1970-01-01T00:00:00Z"`. This seems unlikely.

yaseminbridges:
> This looks good. Only one comment, every dateOfBirth is `"dateOfBirth": "1970-01-01T00:00:00Z"`. This seems unlikely.

I noticed this as well. I just checked the v1 phenopackets and this wasn't there (no details on the dateOfBirth at all) so I think it was done with the `phenopacket-tools` conversion as I didn't add it. I'll remove this field

julesjacobsen:
If this is the case, can you file a bug report for this in the repo, please.

yaseminbridges:
> If this is the case, can you file a bug report for this in the repo, please.

Removed the `dateOfBirth` field and reported the bug in the `phenopacket-tools` repo

julesjacobsen (approved):

PR #109, opened 2023-04-18
108 add missing fields to resources in the metadata

yaseminbridges:
Added missing `version` and `iri_prefix` to OMIM resource  and `iri_prefix` to NCBITAXON resource in phenopackets


julesjacobsen:
NCBITaxon doesn't resolve. Try using [Identifiers.org](https://registry.identifiers.org/registry/taxonomy) to find a resource to replace the iriPrefix e.g.

```json
"iriPrefix": "http://purl.obolibrary.org/obo/NCBITAXON_"
```
with
```json
"iriPrefix": "https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id="
```

matentzn:
No please don't, the one I shared is correct, I will tell you more later.. this is crucial to get right! :P sorry

julesjacobsen:
What Nico said :(

@matentzn  Is this an OWL inclusion thing?

matentzn:
@julesjacobsen I think we crossed wires in this dialogue. The IRI resolves - the one I gave to @yaseminbridges :) All good then?

PR #112, opened 2023-04-18
fixing bug in `generate_pheval_result()`

yaseminbridges:
Python was not able to recognise the typing of list contents, i.e., `[RankedPhEvalVariantResult]` and so threw errors when called in the runner implementation for post-processing. Fixed this issue by checking that all the items in the list are that of the same class and then writing the corresponding result

### DIFF
--- a/src/pheval/post_processing/post_processing.py
+++ b/src/pheval/post_processing/post_processing.py
@@ -76,7 +76,7 @@ class SortOrder(Enum):
 
 class ResultSorter:
     def __init__(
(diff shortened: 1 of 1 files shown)


matentzn (commented):
I cant judge this (too out of the loop with python dev these days) but my gut feeling is telling me https://www.w3schools.com/python/ref_func_all.asp but I might be 100% wrong.

yaseminbridges:
> I cant judge this (too out of the loop with python dev these days) but my gut feeling is telling me https://www.w3schools.com/python/ref_func_all.asp but I might be 100% wrong.

Yes you're completely right! Looks cleaner this way as well

PR #115, opened 2023-04-27
changing VCF file format to be expected in file attributes 

yaseminbridges:
In the `PhenopacketUtil` class and `PhenopacketRebuilder` class, where the file format of the VCF file is either being added/checked/file name taken from the path, these methods looked for files where the file attributes matched `VCF` these needed to be changed to `vcf` for consistency with what is added to the phenopacket after the VCF is spiked as well as what is described in the phenopacket schema documentation


souzadevinicius (commented):
LGTM

PR #119, opened 2023-05-02
decode encoded characters from uri path

yaseminbridges:
The URI path written to the phenopacket after spiking the VCF is now decoded before writing. Using the built-in library `urllib` to decode any URL-encoded characters in the path


matentzn (commented):
I have no opinion here. If you are convinced this is needed!

PR #121, opened 2023-05-02
120 strip non numeric characters from the chromosome length in the vcf header

yaseminbridges:
Stripped non-numeric characters from contig length - required when checking that the VCF assembly matches the variant that is being spiked into the VCF. Added more stringent collecting of the chromosome and contig length in the VCF header


tests/input_dir/test_vcf_dir/test_1.vcf
```diff
@@ -25,31 +25,31 @@
 ##INFO=<ID=MQRankSum,Number=1,Type=Float,Description="Z-score From Wilcoxon rank sum test of Alt vs. Ref read mapping qualities">
 ##INFO=<ID=QD,Number=1,Type=Float,Description="Variant Confidence/Quality by Depth">
 ##INFO=<ID=ReadPosRankSum,Number=1,Type=Float,Description="Z-score from Wilcoxon rank sum test of Alt vs. Ref read position bias">
-##contig=<ID=chrM,length=16571,assembly=hg19>
```
matentzn, line 28: How does this change relate to the PR?

matentzn (commented):
LGTM

tests/input_dir/test_vcf_dir/test_1.vcf
yaseminbridges: I altered one of the test vcfs used for the unit tests to see if the error was fixed. When the header was formatted like: `##contig=<ID=chrM,length=16571>` the contig length returned before would be `16571>\n` instead of just `16571`. I didn't see the error prior as I was splitting on the comma, if something followed the `length` it returned it correctly. I just altered one of the test VCFs instead of adding a new one as the other one has the header formatted like: `##contig=<ID=chrM,length={LENGTH},assembly=hg38>` , so I am still testing for both cases

matentzn: As long as you understand it :) Thanks!

julesjacobsen: These would be better tested as a non-file based unit test specifically to check all the edge cases for these fields. This way you're not potentially breaking other tests and you also have better locality of test data and the test itself.

src/pheval/prepare/create_spiked_vcf.py
```diff
@@ -123,11 +124,15 @@ def parse_assembly(self) -> tuple[str, bool]:
...
+                chromosome = [line for line in line_split if "ID=" in line][0].split("ID=")[1]
                 if "chr" in chromosome:
                     chr_status = True
                     chromosome = chromosome.replace("chr", "")
-                contig_length = line_split[1].split("=")[1]
+                contig_length = re.sub(
+                    "[^0-9]+",
+                    "",
+                    [line for line in line_split if "length=" in line][0].split("length=")[1],
+                )
```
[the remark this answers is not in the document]
yaseminbridges: I would still need to use indices to retrieve the contig length, i.e:

```python
[line for line in line_split if "length=" in line][0]
```

I don't think it would be that easy to use regex to remove the content as other fields can be contained in the contig field (like a URL). In this case, I think it is easier to retrieve the length in this way. However, I can use regex to remove `length=` from the string once retrieved, avoiding using an index in that step

PR #123, opened 2023-05-02
122 add info variable to probandcausativevariant dataclass

yaseminbridges:
Added an `info` field to the `ProbandCausativeVariant` data class. This is implemented in the `causative_variants()` method, when the phenopacket variants are retrieved for spiking into the VCF file. This `info` field is now spiked into the VCF record and structural variants are spiked correctly according to the VCF specification e.g., a deletion variant will contain `N` in the `ref` field and `<DEL>` in the `alt` field. Added relevant tests for any new functionality


julesjacobsen (approved):

souzadevinicius (commented):
LGTM

Can we use [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) to generate our test data?

Issue #117, opened 2023-05-02
Use new rustsim method in semsim process


souzadevinicius:
**Overview**

Create a new Makefile goal that uses the new rustim method for semantic similarity calculation.

PR #124, opened 2023-05-04
Adding new goal to calculate semsim using rust method

souzadevinicius:
New makefile goal creation that uses rustsim method during the semsim calculation.
This is only one illustration that will be used in the future.

Issue #117


Makefile
```diff
@@ -67,6 +69,14 @@ $(RAW_DATA_FOLDER)/hp-mp2.semsim.tsv: $(RAW_DATA_FOLDER)/hp-mp-merged2.db $(RAW_
 	runoak -i $< similarity -p i,p $(TERM) @ $(TERM2) -O csv -o $@
 
 
+results/pheno_rustsim1.semsim1.tsv:
+	test -d results/ || mkdir -p results
+	runoak -i rustsim:sqlite:obo:phenio similarity -p i,p hand @ finger -O csv -o $@
```
matentzn, line 74: We need all the values, not just a handful. How long does this run ti generate the entire table?

souzadevinicius: I tried to, but after more than 3 days we had only a few records :(

matentzn: Please discuss with the rustsim channel in Monarch, we absolutely need to prioritise this!

matentzn (approved):
Make sure this is reflected in #126

PR #128, opened 2023-05-15
fix naming of rank comparison file names

yaseminbridges:
The rank comparison file naming was incorrect due to accessing the same tool config directory for naming the output file. Altered by unpacking the tuple containing the ranks/prioritisation results for the specific run to name the file.

### DIFF
--- a/src/pheval/analyse/generate_summary_outputs.py
+++ b/src/pheval/analyse/generate_summary_outputs.py
@@ -142,29 +142,31 @@ def merge_results(result1: dict, result2: dict) -> dict:
 def generate_gene_rank_comparisons(comparison_ranks: [tuple]) -> None:
     """Generate the gene rank comparison of two result directories."""
     for pair in comparison_ranks:
+        result1, result2 = pair
         merged_results = merge_results(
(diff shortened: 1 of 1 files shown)


src/pheval/analyse/generate_summary_outputs.py
```diff
@@ -142,29 +142,31 @@ def merge_results(result1: dict, result2: dict) -> dict:
 def generate_gene_rank_comparisons(comparison_ranks: [tuple]) -> None:
     """Generate the gene rank comparison of two result directories."""
     for pair in comparison_ranks:
+        result1, result2 = pair
```
matentzn, line 145: Is this pythonic? I understand the idea, but wondering if it's generally recommended (point of failure if pair changes structure).

matentzn (approved):

souzadevinicius (approved):

src/pheval/analyse/generate_summary_outputs.py
yaseminbridges: The most pythonic way is by accessing the index of the element, so I should probably go back to that

matentzn (approved):

souzadevinicius (approved):

PR #135, opened 2023-05-15
134 create dataclasses for pheval disease results

yaseminbridges:
Created dataclasses for a `PhEvalDiseaseResult` and a `RankedPhEvalDiseaseResult`. Altered other result dataclasses to inherit from a base class `PhEvalResult` to help with type hinting in later post-processing functions


src/pheval/post_processing/post_processing.py
```diff
@@ -12,7 +13,12 @@ def calculate_end_pos(variant_start: int, variant_ref: str) -> int:
 
 
 @dataclass
-class PhEvalGeneResult:
+class PhEvalResult:
```
matentzn, line 16: gr8

src/pheval/post_processing/post_processing.py
```diff
@@ -69,27 +75,51 @@ def as_dict(self):
...
+class RankedPhEvalDiseaseResult(PhEvalResult):
+    """PhEval disease result with corresponding rank."""
+
+    pheval_disease_result: PhEvalDiseaseResult
+    rank: int
+
+    def as_dict(self):
+        """Return PhEval disease result as dictionary."""
+        return {
+            "disease_name": self.pheval_disease_result.disease_name,
```
matentzn, line 97: Is this the only place in the entire code that knows the key names? or do you find the string "disease_name" elsewhere? 

If the latter, using strings will violates [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle)

src/pheval/post_processing/post_processing.py
```diff
@@ -69,27 +75,51 @@ def as_dict(self):
...
         """Sort results in descending order."""
         return sorted(self.pheval_results, key=operator.attrgetter("score"), reverse=True)
 
-    def _sort_by_increasing_score(self) -> [PhEvalGeneResult] or [PhEvalVariantResult]:
+    def _sort_by_increasing_score(self) -> Type[PhEvalResult]:
         """Sort results in ascending order."""
         return sorted(self.pheval_results, key=operator.attrgetter("score"), reverse=False)
 
-    def sort_pheval_results(self) -> [PhEvalGeneResult] or [PhEvalVariantResult]:
+    def sort_pheval_results(self) -> Type[PhEvalResult]:
```
matentzn, line 122: Never seen the `Type[` thing. What does it do? Saying that the result must be a subclass of PhEvalResult?

src/pheval/post_processing/post_processing.py
yaseminbridges: Yep, it is the only place that knows the key names, it's just for writing out to a .tsv file. When the file is read back in it is read as a data class so no access to dictionary key names

matentzn: Sweet thanks!

souzadevinicius (approved):
LGTM

src/pheval/post_processing/post_processing.py
yaseminbridges: I believe it declares that the result is a class object deriving from PhEvalResult

matentzn: Not so sure.. I would have thought that `Type[MyClass]` does not refer to a concrete subclass of `MyClass`, but rather to the metaclass of `MyClass`. 

Are you sure the return type should not simply be `[PhEvalResult]`?

yaseminbridges: I read into it some more and it looks like you're right!

matentzn: Thanks for checking again, I learned, again, something new :)

PR #143, opened 2023-05-22
fix blank info field in spiked vcf records

yaseminbridges:
Fixed the occurrence of a blank `INFO` field in spiked VCF records - this happened because the absence of a value in the phenopacket is not `None` and is instead an empty string

### DIFF
--- a/src/pheval/prepare/create_spiked_vcf.py
+++ b/src/pheval/prepare/create_spiked_vcf.py
@@ -206,7 +206,7 @@ class VcfSpiker:
             "100",
             "PASS",
             proband_variant_data.info
-            if proband_variant_data.info is not None
+            if proband_variant_data.info != ""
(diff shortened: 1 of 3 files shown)


src/pheval/prepare/create_spiked_vcf.py
```diff
@@ -206,7 +206,7 @@ def construct_variant_entry(self, proband_variant_data: ProbandCausativeVariant)
             "100",
             "PASS",
             proband_variant_data.info
-            if proband_variant_data.info is not None
+            if proband_variant_data.info != ""
```
matentzn, line 209:
```suggestion
            if not proband_variant_data.info
```
This is a bit more pythonic, but does the same as the your suggestion

matentzn (approved):

PR #144, opened 2023-05-24
Removing whitelines from structural blocks

souzadevinicius:
Generates Makefile in a nicer way removing unnecessary white lines using hyphens inside the structural blocks


matentzn (approved):
Very nice!

Issue #145, opened 2023-05-25
Semsim randomization method implementation


souzadevinicius:
### Overview

This issue is part of #12 meta issue and it comprises the SemSim ETL process.
It starts with `semsim_scramble` method that scrambles semantic similarity profile. The scramble factor will determine how much noise will be added to the semantic similarity value (e.g jaccard similarity)

It then randomises the above scores with a degree of the `scramble_factor` and returns a scrambled dataframe.
Semsim scrambled dataframe is processed to be compatible with exomiser mapping table schema.  `semsim_convert` method.

The final step is a SQL file generation that will be ingested into exomiser database.

PR #147, opened 2023-05-25
145 randomised semsim etl

souzadevinicius:
Semsim methods required in Makefile pipeline. Related to #145


src/pheval/cli_pheval_utils.py
```diff
@@ -244,3 +264,37 @@ def create_spiked_vcfs_command(
...
+    help="Path to the semsim file.",
+)
+@click.option(
+    "--output",
+    "-o",
+    required=True,
+    metavar="FILE",
+    help="Path to the semsim.h2.db converted file.",
+)
+@click.option(
```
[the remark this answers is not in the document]
yaseminbridges: Actually I see in the other file that this is a file prefix, I think you should add `type=str` to the click.options

yaseminbridges (commented):
I just had a few comments, mostly related to the click commands and documentation of the parameters, but looks good

yaseminbridges (approved):
LGTM

src/pheval/cli_pheval_utils.py
```diff
@@ -244,3 +269,41 @@ def create_spiked_vcfs_command(
...
+    "--object-prefix",
+    "-b",
+    required=True,
+    metavar="FILE",
+    help="Object Prefix that will be mapped to the database.",
+    type=str,
+)
+def semsim_convert(input: Path, output: Path, subject_prefix: str, object_prefix: str):
```
matentzn, line 307: Either:

Rename to account for the fact that we may need more convert methods in the future (`semsim_convert_exomiser`)

or

generalise method by adding a `-O/--output-format` parameter to indicate the format of the output (e.g. exomiser h2 db)

I prefer the second personally, but either one is fine.

src/pheval/cli_pheval_utils.py
```diff
@@ -244,3 +269,41 @@ def create_spiked_vcfs_command(
...
+    metavar="FILE",
+    help="Path to the semsim file.",
+    type=Path,
+)
+@click.option(
+    "--output",
+    "-o",
+    required=True,
+    metavar="FILE",
+    help="Path to the semsim.h2.db converted file.",
```
matentzn, line 288: See below, either generalise method, or rename it

src/pheval/cli_pheval_utils.py
```diff
@@ -18,10 +19,34 @@
...
     metavar="FILE",
     help="Path to the semantic similarity profile to be scrambled.",
+    type=Path,
+)
+@click.option(
+    "--output",
+    "-O",
+    metavar="FILE",
+    required=True,
+    help="Path where the SQL generated file will be written.",
```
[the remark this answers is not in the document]
matentzn: We should always separate concerns. There are methods for transforming semantic similarity tables, and others for converting them. We should not have methods that do both at the same time - separation of concern.

src/pheval/utils/utils.py
```diff
@@ -0,0 +1,143 @@
...
+    except TypeError as err:
+        info_log.error(self, exc_info=err)
+        return self
+
+
+def semsim_randomisation(
+    input: Path,
+    output: Path,
+    scramble_factor: float = 0.5,
+    columns_to_be_scrambled: List[str] = tuple(["jaccard_similarity"]),
```
matentzn, line 34: I would make this required and not provide a default.

src/pheval/utils/utils.py
```diff
@@ -0,0 +1,143 @@
...
+        pd.DataFrame: [description]
+    """
+    for col in columns_to_be_scrambled:
+        min_num = dataframe[col].min()
+        max_num = dataframe[col].max()
+        dataframe[col] = dataframe[col].apply(rand, args=(min_num, max_num, scramble_factor))
+    return dataframe
+
+
+def semsimconvert(input: Path, output: Path, subject_prefix: str, object_prefix: str):
```
matentzn, line 75: create a general `semsimconvert` method with a "output_format" parameter, and move all the code in this method here to a specific: `semsimconvert_exomiserdb` or similar.

matentzn: Or to keep it like the h2 method: `semsim2exomiserdb`.

tests/test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_semsimconvert(self):
+        semsim_file = "./testdata/semsim/hp-mp.semsim.tsv"
+        output_file = os.path.join(self._temp_dir.name, "test.tsv")
+        semsimconvert(
+            input=semsim_file, output=output_file, object_prefix="HP", subject_prefix="MP"
+        )
+        with open(output_file, "r") as f:
+            lines = f.readlines()
+            self.assertTrue("INSERT" in " ".join(lines))
```
matentzn, line 23: Can we add an assert that ensures that we have "at least X inserts"?

matentzn: maybe just a check that are are more than 200 lines in this file or some such

tests/test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_semsim_randomisation(self):
+        semsim_file = "./testdata/semsim/hp-mp.semsim.tsv"
+        output_file = os.path.join(self._temp_dir.name, "test.sql")
+        semsim_randomisation(input=semsim_file, output=output_file, scramble_factor=1)
+        with open(output_file, "r") as f:
+            lines = f.readlines()
+            self.assertTrue("subject_id" in " ".join(lines))
```
matentzn, line 31: assert that input tables has the same length as output table, and the same exact columns (number and order)

matentzn (commented):
nice progress on this! 

I made some comments here and there, let me know if you have questions!

src/pheval/cli_pheval_utils.py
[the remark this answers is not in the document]
souzadevinicius: You're absolutely right. There was an error in the documentation. A semsim table is indeed written by this method.

src/pheval/utils/utils.py
```diff
@@ -0,0 +1,143 @@
...
+import random
+from pathlib import Path
+from typing import List
+
+import pandas as pd
+
+info_log = logging.getLogger("info")
+
+
+def rand(self: pd.DataFrame, min_num: int, max_num: int, scramble_factor: float) -> float:
```
matentzn, line 12: is "self" a good variable name? it seems this is used to denote methods that belong to a class?

src/pheval/utils/utils.py
matentzn: This still looks like a method with a general name (semsimconvert) and a specific purpose (converting exomiser db tables).. Do you object renaming it?

src/pheval/cli_pheval_utils.py
```diff
@@ -18,10 +19,44 @@
     required=True,
     metavar="FILE",
     help="Path to the semantic similarity profile to be scrambled.",
+    type=Path,
+)
+@click.option(
+    "--output",
+    "-O",
```
matentzn, line 26:
```suggestion
    "-o",
```

See below (other command), make sure not conflict

src/pheval/utils/utils.py
```diff
@@ -0,0 +1,149 @@
...
+            "IC",
+            "SCORE",
+            "LCS_ID",
+            "LCS_TERM",
+        ]
+    ]
+    input_data.insert(0, "MAPPING_ID", None)
+    input_data["SCORE"].replace("None", "NULL", inplace=True)
+    input_data.to_csv(output, index=False, sep="\t")
+    semsim2h2(input_data, f"{output}", subject_prefix, object_prefix)
```
matentzn, line 124: better name parameters like `input=input_data`

matentzn: input_data does not seem to be a path.

matentzn: Why wrap output in fstring? its a path, pass it through unchanged

src/pheval/utils/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        subject_prefix: (str): mapping subject prefix (e.g. HP)
+        object_prefx (str): mapping object prefix (e.g. MP)
+    """
+    with open(output, "w") as file:
+        insert = f"""
+TRUNCATE TABLE EXOMISER.{subject_prefix}_{object_prefix}_MAPPINGS;
+INSERT INTO EXOMISER.{subject_prefix}_{object_prefix}_MAPPINGS
+(MAPPING_ID, {subject_prefix}_ID, {subject_prefix}_TERM, {object_prefix}_ID, {object_prefix}_TERM, SIMJ, IC, SCORE, LCS_ID, LCS_TERM)
+VALUES
+"""  # noqa
```
matentzn, line 141: dont use noqa

src/pheval/utils/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        file.write(insert)
+        for idx, data in input.iterrows():
```
matentzn, line 143: Its strange qc does not catch this, but `input` is a Path according to method signature, and your you so iterrows.

src/pheval/utils/utils.py
```diff
@@ -0,0 +1,149 @@
...
+            values = f"""
+({idx}, '{data[f'{subject_prefix}_ID']}', '{data[f'{subject_prefix}_TERM']}', '{data[f'{object_prefix}_ID']}', '{data[f'{object_prefix}_TERM']}', {data['SIMJ']}, {data['IC']}, {data['SCORE']}, '{data['LCS_ID']}', '{data['LCS_TERM']}')"""  # noqa
```
matentzn, line 145: why noqa

src/pheval/utils/utils.py
```diff
@@ -0,0 +1,149 @@
...
+    Args:
+        df (pd.DataFrame): dataframe records
+        min_num (int): min value from this records
+        max_num (int): max value from this records
+        scramble_factor (float): scramble factor scalar
+    Returns:
+        float: randomized number
+    """
+    try:
+        return df + (random.uniform(min_num, max_num) * scramble_factor)  # noqa
```
matentzn, line 24: why noqa

matentzn (approved):
Excellent work!

PR #148, opened 2023-05-25
146 altering the PhEval result types defined in the post-processing methods

yaseminbridges:
Altered the PhEval result types defined as dataclasses in the post-processing methods - this mainly involved changing the inheritance of the dataclasses to either the base parent class or the 'unranked' version of the result type. This involved refactoring how variables were obtained in the analysis results and how the results were parsed back into a dataclass (avoiding the use of dictionary keys)

### DIFF
--- a/tests/test_post_processing.py
+++ b/tests/test_post_processing.py
@@ -41,24 +41,44 @@ pheval_variant_result = [
 
 class TestRankedPhEvalGeneResult(unittest.TestCase):
     def setUp(self) -> None:
-        self.pheval_gene_result = RankedPhEvalGeneResult(
+        self.pheval_gene_result = PhEvalGeneResult(
             gene_symbol="A4GNT",
             gene_identifier="ENSG00000118017",
             score=0.6529,
-            rank=1,
+        )
+
+    def test_from_gene_result(self):
+        self.assertEqual(
+            RankedPhEvalGeneResult.from_gene_result(self.pheval_gene_result, 1),
+            RankedPhEvalGeneResult(
+                gene_symbol="A4GNT", gene_identifier="ENSG00000118017", score=0.6529, rank=1
+            ),
         )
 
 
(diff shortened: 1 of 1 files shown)


src/pheval/post_processing/post_processing.py
```diff
@@ -156,20 +162,18 @@ def _return_sort_order(sort_order_str: str) -> SortOrder:
...
     return _rank_pheval_result(sorted_pheval_result, sort_order)
 
 
 def _write_pheval_gene_result(
-    ranked_pheval_result: [RankedPhEvalGeneResult], output_dir: Path, tool_result_path: Path
+    ranked_pheval_result: [PhEvalResult], output_dir: Path, tool_result_path: Path
 ) -> None:
     """Write ranked PhEval gene result to tsv."""
-    ranked_result = pd.DataFrame([x.as_dict() for x in ranked_pheval_result])
+    ranked_result = pd.DataFrame([data.__dict__ for data in ranked_pheval_result])
```
matentzn, line 176: Why this change?

matentzn (approved):
If this works as expected this is a great simplification of the code! Nice.

src/pheval/post_processing/post_processing.py
yaseminbridges: I ended up removing the method `as_dict()` method in the class as Jules wanted me to avoid creating those dictionary/dictionary keys manually

souzadevinicius (approved):
LGTM

julesjacobsen (approved):
Looks way more succinct. You could probably wrap the results in a ranked result and just delegate the calls from RankedResult.chr (for example) to Result.chr  by defining RankedResult as an interface implemented by RankedGeneResult and RankedVariantResult, then you can just call the standard constructor, like so RankedGeneResult(GeneResult, rank) and not need a static constructor to copy the input objects fields.
Still, this works and is better than the previous version.

PR #157, opened 2023-06-14
replacing shell pwd by directories.workspace in Makefile j2 template

### LINKED ISSUE #154: Replace `$(shell pwd)` with `{{ directories.workspace }}` in `Makefile.j2`

souzadevinicius:
fixes #154

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ SHELL 					:= bash
 URIBASE					:=	http://purl.obolibrary.org/obo
 TEST_DATA				:=	testdata
 TMP_DATA				:=	data/tmp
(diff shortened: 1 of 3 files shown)


matentzn (approved):
great, thanks. This begs the question whether `Makefile` should even be in version control, but for now, it helps with reviewing, so lets leave it!

BTW, a slightly more elegant solution would have been this:

```
# Top of the Makefile.j2
ROOT_DIR={{ directories.workspace }}
# ....
configurations/{{ config.tool }}-{{ config.version }}-{{ config.configuration }}/config.yaml:
	rm -rf $(ROOT_DIR)/$(shell dirname $@)
# ....
```

This way you can even overwrite the ROOT dir without regenerating the makefile.

PR #163, opened 2023-06-19
162 create pheval disease results directory

yaseminbridges:
Adding property to runner for retrieving the `pheval_disease_results` directory and creating if specified in the`config.yaml`

### DIFF
--- a/tests/input_dir/configs/default/config.yaml
+++ b/tests/input_dir/configs/default/config.yaml
@@ -1,5 +1,5 @@
 tool: defaultrunner
 tool_version: 1.0.0
 phenotype_only: False
-disease_analysis: False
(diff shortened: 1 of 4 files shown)


tests/input_dir/configs/default/config.yaml
```diff
@@ -1,4 +1,5 @@
 tool: defaultrunner
 tool_version: 1.0.0
 phenotype_only: False
```
julesjacobsen, line 3: Change this to 
```yaml
variant_analysis: False
gene_analysis: True
disease_analysis: True
```
i.e. 'phenotype_only' is a weird amalgam of gene and variant...

julesjacobsen (approved):

PR #175, opened 2023-07-12
164 split configyaml phenotype only into gene and variant fields

yaseminbridges:
Split the `phenotype_only` variable in the config to `gene_analysis` and `variant_analysis`


matentzn (approved):
I think this is clearer.

@souzadevinicius - this PR would have been a good one for the "quick testing pipeline".

Both, can you work together on making sure the changes are reflected in Makefile.j2?

PR #183, opened 2023-08-08
182 generalise parsing methods for reading in pheval result tsv files

yaseminbridges:
Originally there were 3 separate methods for parsing the tsv result to a pheval gene, variant or disease result. I have now generalised this into a single method and instead allow the specification of what result type this needs to be parsed into and moved into its own file separate to the analysis methods


src/pheval/analyse/analysis.py
```diff
@@ -1,4 +1,3 @@
-# #!/usr/bin/python
```
matentzn, line 1: Why was this here in the first place?

src/pheval/analyse/analysis.py
```diff
@@ -16,6 +15,7 @@
     generate_benchmark_gene_output,
     generate_benchmark_variant_output,
 )
+from pheval.analyse.parse_pheval_result import _read_standardised_result, parse_pheval_result
```
matentzn, line 18: Is it a good idea to import a `private` `_...` method to another file?

src/pheval/analyse/parse_pheval_result.py
```diff
@@ -0,0 +1,15 @@
+from pathlib import Path
+
+import pandas as pd
+
+from pheval.post_processing.post_processing import PhEvalResult
+
+
+def _read_standardised_result(standardised_result_path: Path) -> dict:
```
matentzn, line 8: This does not seem to be a private method so why the underscore?

matentzn (commented):
Looks great, just a tiny suggestion!

src/pheval/analyse/analysis.py
yaseminbridges: Definitely some type of mistake on my end, very likely to have happened at the beginning of the project

matentzn (approved):
Great!!

julesjacobsen (approved):
Nice

PR #185, opened 2023-08-09
184 generalise methods for generating plots

yaseminbridges:
Generalised methods used for generating plots. There was a lot of repetition here for generating the gene/variant/disease plots


matentzn (approved):
Looks cool.

julesjacobsen (approved):

PR #194, opened 2023-09-12
191 calculate percentages in the top top3 top5 from total number in corpus

yaseminbridges:
Calculating the summary stats from the total number in the corpus rather than the number found

### DIFF
--- a/tests/test_generate_plots.py
+++ b/tests/test_generate_plots.py
@@ -17,10 +17,10 @@ class TestPlotGenerator(unittest.TestCase):
                 rank_stats=RankStats(
                     top=1,
                     top3=2,
-                    top5=3,
-                    top10=9,
-                    found=20,
-                    total=30,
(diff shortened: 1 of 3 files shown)


tests/test_generate_plots.py
```diff
@@ -61,10 +61,10 @@ def test__generate_stacked_bar_plot_data(self):
...
                 {
                     "2-3": 5.0,
-                    "4-5": 5.0,
-                    "6-10": 30.0,
-                    ">10": 21.66666666666667,
-                    "FO/NP": 33.33333333333333,
+                    "4-5": 10.0,
+                    "6-10": 20.0,
+                    ">10": 10.0,
+                    "FO/NP": 50.0,
```
julesjacobsen, line 67: Is this going to be changed to 'Missing' or something more obvious at some stage?

PR #196, opened 2023-09-12
195 move click commands out of analysispy and move into pheval utils

yaseminbridges:
Moved click commands for benchmarking runs out of `analysis.py` and into `cli_pheval_utils.py`


matentzn (approved):
LGMT, just moving code around.

julesjacobsen (approved):

PR #198, opened 2023-09-14
197 move gene prioritisation specific methods in analysispy

yaseminbridges:
Moved prioritisation type-specific methods/classes into their own files (Disease/Gene/Variant) as well as some other methods/classes into separated files to avoid cyclic imports. None of the actual methods/classes have changed, just moving around code


julesjacobsen (approved):
Much better having split these into more manageable sizes and focused functionalities.

matentzn (approved):
Looks great!

Issue #186, opened 2023-08-09
Generalise RankComparisonGenerator class


yaseminbridges:
Wait for PR #185 review and merge

PR #200, opened 2023-09-21
193 generalise methods that generate summary outputs

yaseminbridges:
Generalised methods for generating summary outputs (these are the summary outputs from the `benchmark` and `benchmark-comparison` commands.

Quick rundown of what has changed:

1. `analysis.py` -> implemented the new generalised methods in place of the previous prioritisation-specific methods
2. `benchmark_generator.py` -> added a new file which contains dataclasses for keeping default values that are used to generate the outputs (y-labels for plots, plot prefixes and what data to return (disease/gene/variant)
3. `benchmarking_data.py` ->  moved these dataclasses from `generate_plots.py` into this new file
4. `generate_plots.py` ->  refactored `FO/NP` to `Missed`. Implemented new dataclasses from `benchmark_generator.py`
5. `generate_summary_outputs.py` -> Main bulk of generalising methods.
6. `constants.py` -> added constant for rank comparison file outputs (avoids SRP @matentzn found in PR #186)
7.  `test_generate_plots.py` -> refactored `FO/NP` to `Missed`

@julesjacobsen This will be the last generalisation PR, and then I will work on implementing the command for generating the plot


src/pheval/analyse/benchmarking_data.py
```diff
@@ -0,0 +1,34 @@
...
+        """Return gene prioritisation analysis results."""
+        return self.gene_prioritisation
+
+    def return_variant(self) -> AnalysisResults:
+        """Return variant prioritisation analysis results."""
+        return self.variant_prioritisation
+
+    def return_disease(self) -> AnalysisResults:
+        """Return disease prioritisation analysis results."""
+        return self.disease_prioritisation
```
julesjacobsen, line 34: Is the `TrackRunPrioritisation` dataclass actually necessary anymore? You have three types of `BenchmarkPrioritisationOutputGenerator` (can't this just be named `BenchmarkOutputGenerator`?) which explicitly calls a single method on the `TrackRunPrioritisation` instance, which returns one of three instances of `AnalysisResults`. Calling the wrong trackRunPrioritisation method will return `None` and the method called is dependent on the type of `BenchmarkOutputGenerator`, so why not just provide the correct `AnalysisResults` instance directly to the `BenchmarkPrioritisationOutputGenerator`  and be done with it?

src/pheval/analyse/benchmark_generator.py
```diff
@@ -0,0 +1,57 @@
...
+    DISEASE_PLOT_Y_LABEL,
+    GENE_PLOT_FILE_PREFIX,
+    GENE_PLOT_Y_LABEL,
+    VARIANT_PLOT_FILE_PREFIX,
+    VARIANT_PLOT_Y_LABEL,
+)
+
+
+@dataclass
+class BenchmarkPrioritisationOutputGenerator:
```
julesjacobsen, line 16: Rename this to `BenchmarkOutputGenerator`? the 'Prioritisation' bit doesn't make sense to me here.

src/pheval/analyse/generate_summary_outputs.py
```diff
@@ -96,53 +91,22 @@ def close(self) -> None:
             print("Error closing ", self.file)
 
 
-def generate_benchmark_gene_output(
-    prioritisation_data: TrackRunPrioritisation, plot_type: str
+def generate_benchmark_output(
+    prioritisation_data: TrackRunPrioritisation,
+    plot_type: str,
+    benchmark_generator: BenchmarkPrioritisationOutputGenerator,
```
julesjacobsen, line 97: The `prioritisation_data` is only used by the `benchmark_generator` so why not just provide the specific `AnalysisResults` directly to the `BenchmarkPrioritisationOutputGenerator` or in this method rather than having a useless `TrackRunPrioritisation` class midway?

julesjacobsen (changes requested):
This is massively improved, but I think you can go one stem further and remove the `TrackRunPrioritisation` class from this and provide the correct `AnalysisResults` directly to the `BenchmarkPrioritisationOutputGenerator` which will make the intention of the code a lot clearer and also prevent passing in an unintentional `None` from a `TrackRunPrioritisation` to a `BenchmarkPrioritisationOutputGenerator`. 

I also think these names need to be made clearer as they are far from self-evident. I think better names would be `TrackRunPrioritisation` -> `BenchmarkRun` and `BenchmarkRunOutputGenerator`. This should hopefully convey a sense that they have some relationship to each other and that they are modelling a benchmark run `TrackRunPrioritisation` doesn't give me any sense about what it is or how it should be used.

yaseminbridges:
@julesjacobsen This should hopefully address all the changes you requested now

src/pheval/analyse/benchmark_generator.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+
+@dataclass
+class BenchmarkRunOutputGenerator:
+    """Base class for recording data required for generating benchmarking outputs."""
+
+    prioritisation_type_file_prefix: str
+    y_label: str
+    benchmark_function: Callable[
```
julesjacobsen, line 27: Rename `benchmark_function` as `generate_benchmark_run`? Isn't the result of the `Callable` a `BenchmarkRun`? Although perhaps `BenchmarkRun` should be called `BenchmarkRunResults` which fits the class documentation `Analysis results for a run`, in which case the method would be better called  `generate_benchmark_run_results` as this is what is does.

src/pheval/analyse/benchmarking_data.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
+from pathlib import Path
+
+from pheval.analyse.rank_stats import RankStats
+
+
+@dataclass
+class BenchmarkRun:
```
julesjacobsen, line 8: As with previous comment  - rename to `BenchmarkRunResults`?

src/pheval/analyse/analysis.py
```diff
@@ -105,44 +45,60 @@ def benchmark_directory(
...
+            results_dir_and_input=results_dir_and_input,
+            score_order=score_order,
+            output_prefix=output_prefix,
+            threshold=threshold,
+            plot_type=plot_type,
+            benchmark_generator=DiseaseBenchmarkRunOutputGenerator(),
+        )
+
+
+def run_benchmark_comparison(
```
julesjacobsen, line 78: Should `run_benchmark_comparison` and `run_benchmark` be a 'private' method? i.e. `_run_benchmark_comparison` and `_run_benchmark` as the names `benchmark` and `benchmark_run_comparisons` are somewhat ambiguous when compared with `run_benchmark_comparison` and `run_benchmark`  and it's not immediately clear which methods should be being called externally.

src/pheval/analyse/generate_summary_outputs.py
julesjacobsen, line 1: Much clearer!

src/pheval/analyse/benchmarking_data.py
julesjacobsen: Also, can't this file and `benchmark_generator.py` be merged into a single `benchmark.py` module?

src/pheval/analyse/rank_stats_writer.py
julesjacobsen, line 1: Merge this file into `rank_stats.py`. I generally like the one file, one class convention Java uses, but here I think they should be grouped into a single `rank_stats` module. It's not going to be too large and groups all the `rank_stats` stuff together.

julesjacobsen (changes requested):
Apart from the `BenchmarkRun` and `benchmark_function` names and some minor package moving about this looks great.

src/pheval/analyse/benchmarking_data.py
yaseminbridges: I can't merge it into a single file as there will be a circular import

julesjacobsen: Where is the circular dependency?

yaseminbridges: The `gene_prioritisation_analysis.py`, `variant_prioritisation_analysis.py` and `disease_prioritisation_analysis.py` import `BenchmarkRunResults` and the benchmark generator imports from each of these files

julesjacobsen (approved):

PR #203, opened 2023-10-09
change variant separators from `_` to `-`

yaseminbridges:
Changed variant separators from `_` to `-` when reporting variant results in the benchmarking


matentzn (approved):
I dont have any context but looks straight forward

PR #209, opened 2023-11-21
208 update docs

yaseminbridges:
Updated the documentation for setting up a PhEval plugin. Added a section for PhEval "helper" methods which would be useful to people setting up a runner implementation


matentzn (approved):
Awesome!!!

PR #216, opened 2023-12-05
213 refactor docstrings in phenopacket utilspy

yaseminbridges:
Refactored docstrings in `phenopacket_utils.py` to Google docstrings.

### DIFF
--- a/src/pheval/utils/phenopacket_utils.py
+++ b/src/pheval/utils/phenopacket_utils.py
@@ -53,9 +53,9 @@ class GenomicVariant:
         chrom (str): The chromosome position of the variant recommended to be provided in the following format.
         This includes numerical designations from 1 to 22 representing autosomal chromosomes,
         as well as the sex chromosomes X and Y, and the mitochondrial chromosome MT.
-        pos (int): Position of the variant
-        ref (str): Reference allele
-        alt (str): Alternate allele
+        pos (int): Position of the variant following VCF convention.
+        ref (str): Reference allele following VCF convention.
+        alt (str): Alternate allele following VCF convention.
     """
 
     chrom: str
--- a/src/pheval/utils/phenopacket_utils.py
+++ b/src/pheval/utils/phenopacket_utils.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 from copy import copy
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Union
+from typing import List, Union
 
 import pandas as pd
 from google.protobuf.json_format import MessageToJson, Parse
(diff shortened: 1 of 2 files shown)


src/pheval/utils/phenopacket_utils.py
```diff
@@ -37,6 +46,18 @@ def __str__(self):
...
+    """
+    Represents a genomic variant.
+
+    Args:
+        chrom (str): The chromosome position of the variant recommended to be provided in the following format.
+        This includes numerical designations from 1 to 22 representing autosomal chromosomes,
+        as well as the sex chromosomes X and Y, and the mitochondrial chromosome MT.
+        pos (int): Position of the variant
+        ref (str): Reference allele
+        alt (str): Alternate allele
```
julesjacobsen, line 58: Is this a VCF-style pos, ref and alt? I'm going to assume so as there are a lot of combinations of these for representing variations. e.g. 1-start or 0-start (VCF uses 1-start), ref and alt alleles can be trimmed or untrimmed (use trimmed), trimming can be left or right (VCF uses left-trimmed), the ref and alt can either be present or absent, depending on how indels are represented. 

e.g. Clinvar variant [13294](https://www.ncbi.nlm.nih.gov/clinvar/variation/13294/) in VCF `10	123256215	.	T	G` a pathogenic missense variant (GRCh37 VCF coordinates - 1-based, positive strand) can be represented in SPDI and HGVS as shown below. It gets more complicated for indels:

Variant type | VCF | SPDI | HGVS
|--------------|-----|-------|-------|
SNP | 10	123256215	.	T	G | NC_000010.10:123256214:T:G | NC_000010.10:g.123256215T>G
DEL | 10	123256215	.	TC	T | NC_000010.10:123256215:C: | NC_000010.10:g.123256216del
INS | 10	123256215	.	T	TA | NC_000010.10:123256215::A | NC_000010.10:g.123256215_123256216insA 

To keep it clear and simple - just refer to the VCF spec.

yaseminbridges: Sorry I’m confused by what you mean here. This is for extracting the data from the PhEval TSV results and from the Phenopacket variant for comparison in the analysis. Is this not over complicating things?

julesjacobsen: I was taking this at face value - representing variants is something people have spent a huge number of hours (careers even?) debating, so just state that these follow VCF convention as tools will need to convert any output data they have into this format. If not stated, you have literally no idea what people will stuff in those fields.

yaseminbridges: Got it! Done

PR #222, opened 2023-12-05
Refactor docstrings in `create_spiked_vcf.py` to Google docstrings

yaseminbridges:
refactoring docstrings to Google docstrings

### DIFF
--- a/src/pheval/prepare/create_spiked_vcf.py
+++ b/src/pheval/prepare/create_spiked_vcf.py
@@ -6,6 +6,7 @@ import urllib.parse
 from copy import copy
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Union
 
 from phenopackets import Family, File, Phenopacket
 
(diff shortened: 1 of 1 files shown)


src/pheval/prepare/create_spiked_vcf.py
```diff
@@ -173,20 +246,42 @@ def check_variant_assembly(
...
+        Construct variant entries.
+
+        Args:
+            proband_variant_data (ProbandCausativeVariant): Data for the proband variant.
+
+        Returns:
+            List[str]: Constructed variant entry as a list of strings.
+        """
         genotype_codes = {
             "hemizygous": "0/1",
```
julesjacobsen, line 286: I think you mean 'heterozygous' here. hemizygous is for single gene copies in a diploid organism. e.g. for Males in non-PAR X. See also https://bioinformatics.stackexchange.com/questions/14356/how-is-the-gt-field-in-a-vcf-file-defined

yaseminbridges: No, I mean hemizygous. There are some phenopackets in the LIRICAL corpus that are hemizygous. I had checked with you at the time if this is the correct genotype code and you approved

julesjacobsen: Ohhh! I see - it's a lookup for converting things. Sorry, I didn't check the context.

PR #227, opened 2023-12-07
changing default plot type to cumulative bar

yaseminbridges:
Changed the default plot type that comes out the benchmarking command to the cumulative bar. 

The labels on the plot will show 'Top, Top3, Top5, Top10'


matentzn (approved):
Big YES

PR #262, opened 2023-12-15
Merge docs into main

yaseminbridges:
Add @souzadevinicius docs files to main - getting the two branches on par so that we can delete the docs branch

Changed the branch for the docs deployment to the `main` branch


yaseminbridges:
@souzadevinicius will need to review these files as he produced them to make sure that they are good to go to merge into the `main` branch

resources/custom.Makefile
souzadevinicius, line 1: Why percent sign was removed?

souzadevinicius (approved):
Just one quick question

resources/custom.Makefile
yaseminbridges: I am not sure, I didn't touch these files. It must be something you have previously changed in the main branch that doesn't match the docs. Please change it if it's not correct

souzadevinicius (approved):

mkdocs.yml
```diff
@@ -43,6 +44,9 @@ watch:
 
 nav:
   - "index.md"
+  - Pipeline:
+    - "exomiser_pipeline.md"
```
matentzn, line 48: Doesn't this have to be  key value pairs? How would mkdocs know otherwise how to render the link?

souzadevinicius: The link will be based on the file's name and the nav title is the file header content.

matentzn (approved):

PR #267, opened 2023-12-22
Update PhEval dependency installation instruction

yaseminbridges:
For now, for developers who want to create a PhEval plugin they should add the PhEval dependency with poetry with the GitHub repo link. This will change once there is a solid release for PhEval on PyPi


yaseminbridges:
Released updated PhEval on PyPi

PR #268, opened 2023-12-22
Updating docs for developing a pheval plugin with notes about `cruft` installation

yaseminbridges:
Adding notes for installing the latest version of cruft to avoid project layout naming issues


souzadevinicius (approved):

matentzn (approved):
For small PRs like this, feel free to just loop in @souzadevinicius 

I don't mind to be added, but more often then not, I may not respond! And you certainly don't need to wait for my review!

PR #271, opened 2024-01-12
Handle empty results

yaseminbridges:
Change `pass` to `return` statement in function. Before, if the result was empty it would continue to try and generate a result - the `return` will exit out of the function to avoid this happening


souzadevinicius (approved):
LGTM

PR #276, opened 2024-01-18
275 calculate MAP@k

yaseminbridges:
Addition of the calculation of Mean Average Precision @ k (MAP@k).

MAP@k evaluates the average Precision at all relevant ranks within the list of top K recommendations. This measure differs from Precision@k as it accounts for the quality of the ranking. 

For instance, when evaluating precision at k (precision@k), the resulting score remains unchanged whether the ranks of the relevant entities are positioned at rank 2 or rank 3. In contrast, Mean Average Precision at k (MAP@k) yields a higher score if the rank of the relevant entity is at rank 2 compared to rank 3.

The main changes that happened to implement this:

1. The addition of a `relevant_result_ranks` variable in the `RankStats` class that holds all relevant ranks. Relevant ranks are the ranks at which the known entities (gene/variant/disease) are found in the results. This was added as knowing how many are in the top n results is not enough information to calculate the results. 
2. Within the `gene_prioritisation_analysis.py`, `disease_prioritisation_analysis`, and `variant_prioritisation_analysis.py` adding the recording of the relevant ranks and adding this to the `RankStats.relevant_result_ranks` variable.


src/pheval/analyse/rank_stats.py
```diff
@@ -158,6 +160,55 @@ def return_mean_reciprocal_rank(self) -> float:
...
+                   A higher value indicates better precision in the top-k predictions.
+        """
+        try:
+            return (1 / number_of_relevant_entities_at_k) * precision_at_k
+        except ZeroDivisionError:
+            return 0
+
+    def mean_average_precision_at_k(self, k: int) -> float:
+        """
+        Calculate the Mean Average Precision at k.
```
matentzn, line 188: Arent the words "Mean" and "average" the same? Above you just say "average" in the private method. Maybe "median average" is a thing, I wouldn't know :D

matentzn (approved):
Looks great!

src/pheval/analyse/rank_stats.py
yaseminbridges: Yep, mean and average are the same, it took me a while to understand the calculation because of its name. The private method first calculates the average precision for a single case in the corpora for the run and then this `mean_average_precision_at_k` calculates the mean for all cases in the corpora for a single run

matentzn: Sure, in this case, why not just name `mean_precision_at_k` - I am not bothered either way :P

julesjacobsen: Keep it as MAP as this is what its commonly called.

PR #280, opened 2024-01-22
279 calculate ndcg@k

yaseminbridges:
Added the calculation of Normalised Discounted Cumulative Gain (NDCG) @ k.

NDCG measures the effectiveness of a ranking by considering the relevance of the results and the order of the items.

For this, I have labelled the known entity as the most relevant (3) and all other results as not relevant (0).

Note: #276 should be reviewed and merged first as this branch is a sub branch for that one. Any changes requested and implemented will need to be reflected here.

PR #278, opened 2024-01-19
277 calculate f scorek

yaseminbridges:

PR #285, opened 2024-01-23
281 implement the functionality to record binary classification categories

yaseminbridges:
Implemented the functionality to record binary classification categories:

True positive -> The known variant/gene/disease ranked 1 in the results.
False positive -> Any other variant/gene/disease ranked 1 in the results.
True negative -> Any other variant/gene/disease ranked at a position other than rank 1.
False negative -> The known variant/gene/disease ranked at a position other than rank 1.

The main changes that were made to implement this:

- Added a `BinaryClassificationStats` class that stores the counts for the binary classification categories.
- Implemented the classification to record the counts for the categories for gene, variant, and disease prioritisation.
- Added tests for this

Note:

PRs #274, #276, #278, #280 should be reviewed and merged first


souzadevinicius (approved):

tests/test_analysis.py
```diff
@@ -1421,3 +1509,9 @@ def test_assess_disease_prioritisation_threshold_meets_cutoff(self):
                 }
             },
         )
+        self.assertEqual(
+            self.binary_classification_stats,
+            BinaryClassificationStats(
+                true_positives=1, true_negatives=3, false_positives=0, false_negatives=0
```
julesjacobsen, line 1515: Are these RankStats correct?
```
RankStats(
    top=1,
    top3=1,
    top5=1,
    top10=1,
    found=1,
    total=1,
    reciprocal_ranks=[1.0],
    relevant_result_ranks=[[1]],
),

true_positives=1, true_negatives=3, false_positives=0, false_negatives=0
```

This looks like found and total should be 4, not 1?

tests/test_binary_classification_stats.py
```diff
@@ -0,0 +1,104 @@
...
+                true_positives=0, true_negatives=4, false_positives=1, false_negatives=0
+            ),
+        )
+
+    def test_add_classification(self):
+        self.binary_classification_stats.add_classification(disease_results, [1, 2])
+        self.assertEqual(
+            self.binary_classification_stats,
+            BinaryClassificationStats(
+                true_positives=1, true_negatives=2, false_positives=0, false_negatives=1
```
julesjacobsen, line 102: Given 2 results, with ranks 1, 2, assuming the rank 1 is a TP, then 2 must be a TN and then the FP and FN should be 0, no?

julesjacobsen (commented):
I might be getting confused here, but are you sure you're not double-counting the values?

tests/test_analysis.py
yaseminbridges: This is correct. Found and total should be 1 as the found and total point to the number of relevant entities found. But the TP, TN, FP, FN points to all the results, relevant or not.

tests/test_binary_classification_stats.py
yaseminbridges: No, for this I imagined that there are 2 variants linked for a sample and one of the variants is found ranked 1 in the results and the other is ranked 2. This would mean there is 1 TP and 1 FN. There are 2 true negatives as these are the other results that are not relevant.

PR #292, opened 2024-01-30
Removing H2 jar as internal dependency. Fixes #291

### LINKED ISSUE #291: Remove H2 jar as an internal dependency
Moving the h2 jar dependency out of the pheval repository. This is because during the poetry installation, the library path changes to a different structure. As a result, I have created a new parameter called "pheval-utils semsim-to-exomiserdb" with the option "-h2-jar". Users should specify the location of the h2 jar file using this parameter.

souzadevinicius:
Fixes #291


matentzn (approved):
Looks like a good move to me, but better have a @yaseminbridges review for details

julesjacobsen:
Hardcoding the H2 file is going to break on the new version as the database file format is not stable. The best place to get the dependency from would be the exomiser-cli-/lib folder as this is the version used to create and read the files.

julesjacobsen (changes requested):
See my comment - in the PR

yaseminbridges (approved):
LGTM

julesjacobsen:
@souzadevinicius Sorry, I've completely forgotten about this! I wonder if things have changed a bit more since this was first looked and and we need to revisit what the point of the methods using this file is. The intention of these are to be able to load arbitrary semsim data into an Exomiser database to be able to run it in a pheval benchmark. This is fine, but it doesn't belong in this pheval repo and needs to be removed somewhere else. There should be no application-specific code in this repo as this is essentially a library which other runners/projects need to import and none of them will need this code. Can you please move it into another repo.

PR #294, opened 2024-02-01
check for no parent in `convert_patient_terms_to_parent`

yaseminbridges:
Implementing a fix for an `IndexError` seen when there is no parents found for a given HPO term

### DIFF
--- a/src/pheval/prepare/create_noisy_phenopackets.py
+++ b/src/pheval/prepare/create_noisy_phenopackets.py
@@ -121,15 +121,16 @@ class HpoRandomiser:
         parent_terms = []
         for term in hpo_terms_to_be_changed:
             try:
-                parent_terms.append(
-                    self.retrieve_hpo_term(self.hpo_ontology.hierarchical_parents(term.type.id)[0])
-                )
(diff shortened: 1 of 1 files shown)


src/pheval/prepare/create_noisy_phenopackets.py
```diff
@@ -121,15 +121,16 @@ def convert_patient_terms_to_parent(
         parent_terms = []
         for term in hpo_terms_to_be_changed:
             try:
-                parent_terms.append(
-                    self.retrieve_hpo_term(self.hpo_ontology.hierarchical_parents(term.type.id)[0])
-                )
+                parents = self.hpo_ontology.hierarchical_parents(term.type.id)
+                parent_terms.append(self.retrieve_hpo_term(random.choice(parents)))
```
julesjacobsen, line 125: What could throw the error here? Is it being thrown because the term is obsolete? If so, why not just check if the term is obsolete first,  handle that and then just follow the logic you have in the exception. This way, no exception checking required.

souzadevinicius (approved):

PR #297, opened 2024-02-09
Move legend in stats plot generation

yaseminbridges:
Moved the legend position in the stats plots to the bottom of the plot


julesjacobsen:
Have you got an example of what the change looks like?

PR #301, opened 2024-02-28
300 change rank decrease to rank change in rank comparison summary

yaseminbridges:
Changed the `rank_decrease` column to `rank_change` in the rank comparison summary file. In cases where a result is not found this has been replaced by a `GAINED` or `LOST` label.

e.g.,

```tsv
	Phenopacket	Gene	exomiser-13.3.0-2109/lirical-default/pheval_gene_results	GPT-4_turbo/complex_prompt/pheval_gene_results	rank_change
1	Abdul_Wahab-2016-GCDH-Patient_5.json	GCDH	1	1.0	0.0
2	Ajmal-2013-BBS1-IV-5_family_A.json	BBS1	3	0.0	LOST
3	Al-Dosari-2010-TFAP2A-10-year-old_girl.json	TFAP2A	0	66	GAINED
4	Al-Hashmi-2018-SNX14-IV-1.json	SNX14	9	6	-3
```

### DIFF
--- a/tests/test_generate_summary_outputs.py
+++ b/tests/test_generate_summary_outputs.py
@@ -102,7 +102,7 @@ class TestRankComparisonGenerator(unittest.TestCase):
                     "Gene": "GCDH",
                     "/path/to/results_directory1": 1,
                     "/path/to/results_directory2": 5,
-                    "rank_change": 4,
+                    "rank_change": -4,
                 }
             ]
         )
@@ -117,7 +117,7 @@ class TestRankComparisonGenerator(unittest.TestCase):
                     "Variant": "3-12563453454-C-T",
                     "/path/to/results_directory1": 9,
                     "/path/to/results_directory2": 3,
-                    "rank_change": -6,
+                    "rank_change": 6,
                 }
             ]
         )
--- a/tests/test_generate_summary_outputs.py
+++ b/tests/test_generate_summary_outputs.py
@@ -102,7 +102,7 @@ class TestRankComparisonGenerator(unittest.TestCase):
                     "Gene": "GCDH",
                     "/path/to/results_directory1": 1,
                     "/path/to/results_directory2": 5,
-                    "rank_decrease": 4,
+                    "rank_change": 4,
                 }
             ]
         )
@@ -117,7 +117,7 @@ class TestRankComparisonGenerator(unittest.TestCase):
                     "Variant": "3-12563453454-C-T",
                     "/path/to/results_directory1": 9,
                     "/path/to/results_directory2": 3,
-                    "rank_decrease": -6,
+                    "rank_change": -6,
                 }
             ]
         )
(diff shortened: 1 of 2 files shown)


tests/test_generate_summary_outputs.py
```diff
@@ -102,7 +102,7 @@ def test_calculate_gene_rank_difference(self):
                     "Gene": "GCDH",
                     "/path/to/results_directory1": 1,
                     "/path/to/results_directory2": 5,
-                    "rank_decrease": 4,
+                    "rank_change": 4,
```
julesjacobsen, line 105: Isn't a rank decrease of 4 equal to a rank change of -4?

tests/test_generate_summary_outputs.py
```diff
@@ -117,7 +117,7 @@ def test_calculate_variant_rank_difference(self):
                     "Variant": "3-12563453454-C-T",
                     "/path/to/results_directory1": 9,
                     "/path/to/results_directory2": 3,
-                    "rank_decrease": -6,
+                    "rank_change": -6,
```
julesjacobsen, line 120: Isn't a rank decrease of -6 equal to a rank change of 6?

julesjacobsen (changes requested):
Can you double-check that the numbers are correct in your tests? A negative rank decrease is a positive rank change. Conversely a positive rank decrease is a negative rank change.

One other thing - changing an integer column to a mostly integer, but with a few `INCREASE` and `DECREASE` values is going to affect what downstream? How are these fields used in pheval? Are they only for human use?

tests/test_generate_summary_outputs.py
yaseminbridges: If we are comparing `results_directory1` to `results_directory2` using `results_directory1` as the baseline, hasn't the rank increased by 4 ranks?

tests/test_generate_summary_outputs.py
yaseminbridges: If we are comparing `results_directory1` to `results_directory2` using `results_directory1` as the baseline, hasn't the rank decreased by 6 ranks?

yaseminbridges:
> Can you double-check that the numbers are correct in your tests? A negative rank decrease is a positive rank change. Conversely a positive rank decrease is a negative rank change.
> 
> One other thing - changing an integer column to a mostly integer, but with a few `INCREASE` and `DECREASE` values is going to affect what downstream? How are these fields used in pheval? Are they only for human use?

I'm getting a bit confused. I looked at the rank change as to how the rank has changed from the first results directory to the second directory. So if the rank has improved in the second run then a negative rank change is reported as the actual rank has decreased.

The fields and results in this file is only for human use and won't affect anything downstream

tests/test_generate_summary_outputs.py
yaseminbridges: Changed now!

PR #326, opened 2024-05-02
325 alter ranking of results penalising results found at the same rank

yaseminbridges:
Altered the ranking of the post-processed results so equal ranks are now penalised.

Before:

1,1,3

Now

2,2,3

### DIFF
--- a/src/pheval/post_processing/post_processing.py
+++ b/src/pheval/post_processing/post_processing.py
@@ -227,68 +227,7 @@ class ResultSorter:
         )
 
 
-class ScoreRanker:
-    """
-    Class for ranking scores based on a given sort order
-
-    Attributes:
-       rank (int): Represents the current rank, initialised with 0
-       current_score (float): Represents the current score, initialised with positive infinity (float("inf"))
-       count (int): Used for counting, initialised with 0
-    """
-
-    rank: int = 0
-    current_score: float = float("inf")
-    count: int = 0
-
-    def __init__(self, sort_order: SortOrder):
-        """
-        Initialise ScoreRanker
-
-        Args:
-            sort_order (SortOrder): Sorting order to be applied
-        """
-        self.sort_order = sort_order
-
-    def _check_rank_order(self, round_score: float) -> None:
-        """
-        Check if the results are correctly ordered
-
-        Args:
-            round_score (float): Score to be checked against the current score
-
-        Raises:
-            ValueError: If results are not correctly sorted.
-        """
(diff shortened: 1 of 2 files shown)


src/pheval/post_processing/post_processing.py
```diff
@@ -227,68 +227,7 @@ def sort_pheval_results(self) -> [PhEvalResult]:
         )
 
 
-class ScoreRanker:
```
souzadevinicius, line 230: Has the class been moved?

tests/test_post_processing.py
```diff
@@ -216,40 +217,40 @@ def test_sort_pheval_results_pvalue(self):
...
-        self.assertEqual(self.score_ranker_descending.rank_scores(0.7342), 1)
-        self.assertEqual(self.score_ranker_descending.rank_scores(0.3452), 2)
-        self.assertEqual(self.score_ranker_descending.rank_scores(0.3452), 2)
-
-    def test_rank_scores_count_increase(self):
-        self.assertEqual(self.score_ranker_descending.rank_scores(0.7342), 1)
-        self.assertEqual(self.score_ranker_descending.rank_scores(0.3452), 2)
-        self.assertEqual(self.score_ranker_descending.rank_scores(0.3452), 2)
-        self.assertEqual(self.score_ranker_descending.rank_scores(0.1234), 4)
+# class TestScoreRanker(unittest.TestCase):
```
souzadevinicius, line 220: Will this test return in the future?

souzadevinicius:
> Altered the ranking of the post-processed results so equal ranks are now penalised.
> 
> Before:
> 
> 1,1,3
> 
> Now
> 
> 2,2,3

Why equal ranks are penalised?

souzadevinicius (approved):

yaseminbridges:
> > Altered the ranking of the post-processed results so equal ranks are now penalised.
> > Before:
> > 1,1,3
> > Now
> > 2,2,3
> 
> Why equal ranks are penalised?

Damian and Jules agreed that this is the better way to rank results. In theory with the current ranking if a tool outputs all genes with the same score and they were all ranked 1 the results would be misleading. If it is penalised we can account for these types of instances.

src/pheval/post_processing/post_processing.py
yaseminbridges: It has been removed.

tests/test_post_processing.py
yaseminbridges: No, as the class has been removed.

PR #334, opened 2024-05-23
333 allow multiple genes/gene identifiers for a single ranked result

yaseminbridges:
Some tools may output alternate gene identifiers/symbols for the same top-ranked variant (AI-MARRVEL as an example).

To allow for this, the post-processing has now allowed for either a single gene symbol/identifiers or a list of gene symbols/identifiers - this avoids breaking the other runners.

The benchmarking code for the gene analysis has also been altered - processing either a list of gene symbols/gene identifiers or a single one to match against the causative gene(s)


souzadevinicius (approved):
LGTM

PR #338, opened 2024-06-04
Add a pipeline test that invokes pheval.template

### LINKED ISSUE #337: Add a pipeline test that invokes pheval.template
Include a test that calls a Makefile configured to run pheval.template with a fake runner

souzadevinicius:
Fixes #337

I apologize if this review seems a bit extensive, but it's necessary because the Makefile pipeline in this project was outdated compared to the one in the [Monarch PhEval repository](https://github.com/monarch-initiative/monarch_pheval).
Basically in this PR, I made changes to the [QC workflow file](https://github.com/monarch-initiative/pheval/pull/338/files#diff-660f01989d3769165300abbe34bb9414ab7b60f2bf78db20ff86a8f50a6c49bf) removing --no-root flag, to install PhEval into the poetry environment. The test itself is located in [tests/test_pipeline_workflow.yaml](https://github.com/monarch-initiative/pheval/pull/338/files#diff-892278afb3c231c2e990f3c446d087749c8ebf55a95ce033baebe4d3a664fccb), it executes "make all" and ensures that pheval run is called and checks the presence of result files.


.github/workflows/qc.yaml
```diff
@@ -77,7 +77,7 @@ jobs:
       #----------------------------------------------
       - name: Install dependencies
         # if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-        run: poetry install --no-interaction --no-root
+        run: poetry install --no-interaction
```
matentzn, line 81: Shouldnt we add an integration test here?

matentzn:
@yaseminbridges we ran this locally and it seems to work good! Can you look this over when you get a minute?

pyproject.toml
```diff
@@ -27,6 +27,7 @@ matplotlib = "^3.7.0"
 pyserde = "^0.9.8"
 polars = "^0.19.15"
 scikit-learn = "^1.4.0"
+pheval-template = "^0.1.1"
```
yaseminbridges, line 30: does this need to be a dependency? Or could it be pip installed into the environment?

yaseminbridges (commented):
LGTM

pyproject.toml
souzadevinicius: No, it doesn't. I will install it with pip in QC workflow.

matentzn: Enough as "dev" dependency?

PR #340, opened 2024-06-12
339 add synthetic corpus

yaseminbridges:
Added synthetic corpus based on HPOA - single phenopacket/patient for each OMIM disease in the annotation.


matentzn (approved):

souzadevinicius (approved):

yaseminbridges:
@matentzn This is technically ready, but are you happy with the 5000+ files being checked into version control?

matentzn:
Ok, shall we instead just attach the synthetic corpus to the pheval github release?

yaseminbridges:
I think that might be the best way to go about it

PR #360, opened 2024-09-19
Added an option to the scramble phenopackets command to allow for a local cached ontology file

yaseminbridges:
Allows the specification of hp.obo/hp.json etc etc to be consumed by OAK and used in the workflow for scrambling phenopackets in place of fetching remotely


julesjacobsen (changes requested):
If you added all the `scramble_phenopackets` functions as methods on the `HpoRandomiser` class then you wouldn't have to change the method signature of all of these as where the ontology was coming from would be dealt with once when constructing the `HpoRandomiser`

PR #368, opened 2024-11-05
363 account for compound het variants

yaseminbridges:
This introduces enhanced ranking functionality to handle compound heterozygous variants in PhEval results through a new `grouping_id` parameter. The grouping_id allows variants that contribute collectively to a phenotype (e.g., compound heterozygous variants within the same gene) to be ranked together (and not penalised), ensuring their combined effect is accurately represented in the analysis. If variants are not compound heterzygous, they will be ranked with the default method.

If a `grouping_id` is not specified the code will **not** break, it will just default to the current method of ranking variants - this is important as not all VGPAs will give an idea of whether a variant is compound het or not and so they can just dismiss this.

### DIFF
--- a/src/pheval/post_processing/post_processing.py
+++ b/src/pheval/post_processing/post_processing.py
@@ -99,14 +99,14 @@ class PhEvalVariantResult(PhEvalResult):
     grouping_id: str = field(default=None)
 
 
-@dataclass(kw_only=True)
+@dataclass
 class RankedPhEvalVariantResult(PhEvalVariantResult):
     """PhEval variant result with corresponding rank
     Args:
         rank (int): The rank for the result entry
     """
 
-    rank: int
+    rank: int = 0
 
     @staticmethod
     def from_variant_result(pheval_variant_result: PhEvalVariantResult, rank: int):
--- a/src/pheval/post_processing/post_processing.py
+++ b/src/pheval/post_processing/post_processing.py
@@ -1,6 +1,6 @@
 import logging
 import operator
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
 from typing import List, Union
@@ -96,9 +96,10 @@ class PhEvalVariantResult(PhEvalResult):
     ref: str
     alt: str
     score: float
+    grouping_id: str = field(default=None)
 
 
-@dataclass
+@dataclass(kw_only=True)
 class RankedPhEvalVariantResult(PhEvalVariantResult):
     """PhEval variant result with corresponding rank
     Args:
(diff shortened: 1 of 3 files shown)


souzadevinicius (approved):
LGTM

tests/test_post_processing.py
```diff
@@ -217,41 +217,145 @@ def test_sort_pheval_results_pvalue(self):
...
+            }
+        )
+        self.assertTrue(
+            self.gene_result_ranker._rank_with_grouping_id(df).equals(
+                pd.DataFrame(
+                    {
+                        "score": [0.9, 0.9, 0.8, 0.7, 0.7, 0.6, 0.5, 0.5, 0.4],
+                        "grouping_id": ["A", "A", "B", "C", "D", "E", "F", "F", "G"],
+                        "min_rank": [1.0, 1.0, 3.0, 5.0, 4.0, 6.0, 7.0, 7.0, 9.0],
+                        "rank": [1.0, 1.0, 3.0, 5.0, 5.0, 6.0, 7.0, 7.0, 9.0],
```
julesjacobsen, line 333:
```
"score":       [0.9, 0.9, 0.8, 0.7, 0.7, 0.6, 0.5, 0.5, 0.4],
"grouping_id": ["A", "A", "B", "C", "D", "E", "F", "F", "G"],
"min_rank":    [1.0, 1.0, 3.0, 5.0, 4.0, 6.0, 7.0, 7.0, 9.0],
"rank":        [1,     1,   2,   4,   4,   5,   6,   6,   7],
```

julesjacobsen:
Have you specified what the default values should be in the documentation and given some examples of their usage for comp het and non-comp het variants?

Have you added a validity check that the grouping ids are unique and where two variants are grouped that they are from the same gene (assuming a monogenic disorder)? Were you going to expand this concept to genes and diseases for di/polygenic disorders or those affected with multiple disorders?

yaseminbridges:
> Have you specified what the default values should be in the documentation and given some examples of their usage for comp het and non-comp het variants?

I have done this in a way that it will be backwards compatible. I didn't want to implement it in a way that this change meant that _all_ runners would need to be updated to implement this. So I guess there are no need for default values, as it will work if none are specified. I've added details in the docs what the `grouping_id` is for and the code will update automatically so that developers will be able to see the usage.

> Have you added a validity check that the grouping ids are unique and where two variants are grouped that they are from the same gene (assuming a monogenic disorder)? Were you going to expand this concept to genes and diseases for di/polygenic disorders or those affected with multiple disorders?

I have not added a validity check, when we discussed this before you said to keep it "open" for developers to decide. The actual implementation (however they decide to classify these variants that should be considered as a single entity) is done in the runners. There may be a tool out there that has some other way of classifying these without using the gene and so I would want to avoid hardcoding this into the PhEval codebase as the result output may not have it and we wanted to keep it to the bare minimum information required to make it easy and "doable" for developers.

The idea was to expand this concept of using a grouping id which is why I've kept it general ish for now.

src/pheval/post_processing/post_processing.py
```diff
@@ -228,26 +229,57 @@ def sort_pheval_results(self) -> [PhEvalResult]:
...
-    Raises:
-        ValueError: If an incompatible PhEval result type is encountered
-    """
-    pheval_result_df = pd.DataFrame([data.__dict__ for data in pheval_result])
-    if sort_order == SortOrder.ASCENDING:
-        pheval_result_df["rank"] = pheval_result_df["score"].rank(method="max", ascending=True)
-    elif sort_order == SortOrder.DESCENDING:
-        pheval_result_df["rank"] = pheval_result_df["score"].rank(method="max", ascending=False)
-    return pheval_result_df
+        if self._has_valid_grouping_id(pheval_result_df):
```
julesjacobsen, line 253: If there are `None`s mixed with grouping_ids in the dataFrame, shouldn't this be an error state? This only checks for the presence of any `None` and then just uses `_rank_without_grouping_id` if there is a `None`. Surely the "grouping_id" column must either be all `None` or all some otherwise the grouping won't work?

yaseminbridges: Happy to go with whatever you think is best practice here. My idea was that if there was `None` mixed I would just default the ranking behaviour ignoring the missing ones as I didn't want the whole process to finish early and continue doing the next ones. But if you think it is best to throw an error then I can change it to that

julesjacobsen: I think it should result in an error as silently ignoring the grouping id isn't a nice thing to do. It needs to be made clear that when a grouping id is used then all results should have one. This isn't made clear enough in the docs.

I'm going to merge this anyway as this will need changing when migrating over to a parquet and polars solution.

PR #392, opened 2025-03-07
Pandas to Polars/Parquet migration

yaseminbridges:
# **Major Migration: Pandas → Polars for Performance and Efficiency**

This PR introduces a **significant** migration from **Pandas to Polars**, improving both **speed** and **RAM efficiency** during post-processing in the `pheval run` command and benchmarking.

⚠️ **Breaking Change**: All runners must be updated to use the new functionalities.

## **Key Project Changes**
- Updated the minimum python version for the project to 3.10 as the `match` functionalities with Enum is not supported in 3.9
- Started to implement improved logging.
## **Key Post-Processing Changes**
- The causative entity is now labelled as **`true_positive`** in the PhEval result file to help with calculating benchmarking statistics.
- PhEval result files have transitioned from **TSV format** to a **compressed Parquet format** for improved performance and storage efficiency.
- The input for helper commands (ranking and writing result files) is now handled using **Polars DataFrames** instead of Pandas. This is where the key change is required from all the runners.
## **Key Benchmarking Changes**
- Consumes the parquet files instead of the TSV format. This now functions as a scan of the entire directory and works with the labelling of the true positive in the file. This allows the separate files for variant/disease/gene prioritisation to be removed as it can all be generalised, resulting in a **massive** cut down of code.
- Uses polars, duckdb, and arrays in favour over pandas and python lists for quicker and more efficient run time.

## 🚀 **Benchmarking Results: A Massive Performance Boost!**
Benchmarked **111 phenopackets** on **Exomiser (phenotype-only) and GADO**. The results speak for themselves:

| **Implementation**  | **Time Taken**  | **Speed Improvement 🚀** |
|---------------------|----------------|--------------------------|
| 🔴 **Old (Pandas-based)**  | ⏱ **41.83 seconds** | ❌ Slow & memory-heavy |
| 🟢 **New (Polars-based)**  | ⏱ **2.09 seconds**  | ✅ **~20x faster!** ⚡ |

### **Key Wins:**
✅ **~20x faster benchmarking**   
✅ **Dramatic reduction in RAM usage**   
✅ **Cleaner and more efficient codebase**


src/pheval/cli.py
julesjacobsen, line 1: I think, in the end, the `pheval run` command should probably be the responsibility of the runner e.g.
 `pheval.exomiser run config.yml`

Then `pheval-util` is really `pheval` e.g. `pheval benchmark config.yml`  or `pheval generate-plots benchmark.duckdb --run-data config.yml`

Then we can get rid of `poetry` and move to `uv` so sanity can prevail.

julesjacobsen (approved):
Quite the change! I couldn't spot anything glaringly wrong, and you showed me that the end-to-end tests produce identical results to the original implementation, so I think all is fine.

PR #416, opened 2025-06-09
367 map to mondo in disease benchmarking

yaseminbridges:
### Disease Identifier Mapping to MONDO

- Added support for mapping disease identifiers (e.g., OMIM, ORPHA) to MONDO IDs during post-processing of disease results.
- Introduced a new column: `mondo_identifier` in the disease results DataFrame.
- This ensures that both the original disease_identifier and its MONDO equivalent are available for benchmarking and analysis.
- ⚠️ Breaking change for the benchmarking output format:  now expect the additional mondo_identifier column. This is not something for runners/developers to be concerned about as it is dealt with in PhEval but it is something important to note if trying to benchmark files generated with PhEval versions <0.6.0.
- Runners are not impacted, as this change applies only to post-processing and result generation.

### New `pheval update` Command

- Added a CLI command: `pheval update`
- Downloads the latest versions of:
    - `mondo.sssom.tsv` from the Monarch Initiative
    - `hgnc_complete_set.txt` from HGNC
- Stores both in the `resources/` directory.
- Writes a `metadata.json` file that tracks the download timestamp for each resource.
- The download timestamp for each resource is written after each `pheval run` to the `results.yaml` in the `output directory`

Additional Notes

- As part of this update, I also updated the `mondo.sssom.tsv` and `hgnc_complete_set.txt` files in the repo.

### DIFF
--- a/src/pheval/post_processing/post_processing.py
+++ b/src/pheval/post_processing/post_processing.py
@@ -45,6 +45,10 @@ def _rank_results(results: pl.DataFrame, sort_order: SortOrder) -> pl.DataFrame:
     """
     sort_descending = True if sort_order == SortOrder.DESCENDING else False
     group_by = []
+    # group by either `grouping_id` or `mondo_identifier` column to implement ranking logic where equal ranks are not
+    # penalised. `grouping_id` and `mondo_identifier` cannot be grouped together as ranking logic fails so
+    # `grouping_id` was removed from the disease results schema.
+    # TODO: allow both `grouping_id` and `mondo_identifier` for disease results to account for blended phenotypes.
     if "grouping_id" in results.columns:
         group_by.append("grouping_id")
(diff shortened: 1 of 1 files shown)


src/pheval/post_processing/post_processing.py
```diff
@@ -41,12 +44,16 @@ def _rank_results(results: pl.DataFrame, sort_order: SortOrder) -> pl.DataFrame:
         pl.DataFrame: The ranked results.
     """
     sort_descending = True if sort_order == SortOrder.DESCENDING else False
-    has_grouping_id = "grouping_id" in results.columns
-    if has_grouping_id:
+    group_by = []
```
julesjacobsen, line 47: Add a TODO and explain why you're doing this. Future you will thank past you.

